PR #6, opened 2019-03-03
Add scale-in cooldown support

etaoins:
This reimplements ASG scale-in cooldown inside the lambda. It takes two parameters which correspond to the existing ASG parameters in the elastic CI stack.

1. `SCALE_IN_COOLDOWN_PERIOD` is the cooldown time between scale in events. This defaults to the existing 5 minutes.

2. `SCALE_IN_ADJUSTMENT` is the maximum adjustment during scale-in events. Unlike the ASG we may scale in less if we calculate that the desired is closer than the adjustment. This defaults to the existing -1.

This cheats a bit by storing `lastScaleInTime` in a global variable. This means we'll forget about our cooldown during a cold start. This should happen fairly infrequently and just make us a bit aggressive about scaling it; it shouldn't affect correctness.

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -17,6 +17,10 @@ import (
 
 var invokeCount = 0
 
+// lastScaleDownTime stores the last time we scaled down the ASG
+// On a cold start this will be reset to Jan 1st, 1970
+var lastScaleInTime time.Time
+
 func main() {
 	if os.Getenv(`DEBUG`) != "" {
 		_, err := Handler(context.Background(), json.RawMessage([]byte{}))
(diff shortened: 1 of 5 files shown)


lox:
This is awesome @etaoins! Somehow I totally missed this PR.

lox:
I wonder how often there will be a cold start on a timed lambda 🤔

etaoins:
I had a timed Lambda running for a few weeks that would cold start about once a day at an unpredictable time. It looks like even if the Lambda is warm its instance will sometimes disappear anyway. That's only a single datapoint that's about a year out of date; I wouldn't be surprised if there are other situations or configurations that could trigger it more often.

I don't think that's too bad here as it will only make a difference if a scale-in is occurring and we will still be limited by the maximum adjustment. My gut instinct would be to try this with local state and if there are some pathological cases it could then be persisted somewhere. I'm just worried about the complexity/reliability of setting up something like a DynamoDB to track this.

PR #9, opened 2019-03-25
Factor running jobs into scaling decisions

### LINKED ISSUE #7: Desired capacity does not account for running jobs
Unless I'm misunderstanding, it appears that autoscaling doesn't not account for currently running jobs. For example, in this situation:

1. One scheduled job
2. One running job
3. One running agent

We think our desired capacity should be 1 when it should actually scale up to 2.

Let me know if I'm understanding this correctly and I'll open a PR.

lox:
Closes #7.

### DIFF
--- a/buildkite/buildkite.go
+++ b/buildkite/buildkite.go
@@ -29,25 +29,9 @@ func NewClient(agentToken string) *Client {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


lox:
Any feels @etaoins?

etaoins:
LGTM!

PR #12, opened 2019-04-28
Replace adjustments with min/max for scale in and out

lox:
Currently we have scale-in-adjustment and cooldown, but when I went to extend this to a scale-out adjustment I realized that there are two distinct use cases, a minimum bound and a maximum bound for each scaling direction.

For scale-in, you might want to slow down scale-in, for which you'd set a `scale-in-max` of say `-2`, which would ensure that the scale-in didn't move by more than `-2`. Alternately, you might (for some reason) want to scale in always by the maximum, so you could set a `scale-in-min` of `-100`. This one is pretty tenuous, but I implemented it to match it's more useful `scale-out-max`.

For scale-out, you might want to have faster scale-out, so you set a `scale-out-min` of 20, which means that you always scale by at least 20 instances. If you set `scale-out-min`, you can control the speed at which scale out runs. 

I also implemented a `disable` switch for scale-out and a cooldown timer if you are so inclined.

@etaoins I'd love your feedback on `scale-in-adjustment` vs `scale-in-min` and `scale-in-max`. Min/max with negative numbers is super confusing but I couldn't figure out how to handle the scale out and keep them consistent 🤔

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -15,11 +15,11 @@ import (
 	"github.com/buildkite/buildkite-agent-scaler/version"
 )
 
-var invokeCount = 0
-
-// lastScaleDownTime stores the last time we scaled down the ASG
-// On a cold start this will be reset to Jan 1st, 1970
-var lastScaleInTime time.Time
+// Stores the last time we scaled in/out in global lambda state
+// On a cold start this will be reset to a zero value
+var (
+	lastScaleIn, lastScaleOut time.Time
+)
 
 func main() {
 	if os.Getenv(`DEBUG`) != "" {
@@ -28,8 +28,6 @@ func main() {
 			log.Fatal(err)
 		}
 	} else {
-		invokeCount = invokeCount + 1
-		log.Printf("Invocation count %d", invokeCount)
 		lambda.Start(Handler)
 	}
 }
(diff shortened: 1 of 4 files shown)


lambda/main.go
```diff
@@ -28,53 +28,85 @@ func main() {
...
+		}
+	}
 
-		if scaleInAdjustment >= 0 {
-			panic(fmt.Sprintf("Scale in adjustment (%d) must be negative", scaleInAdjustment))
+	if v := os.Getenv(`SCALE_IN_MIN`); v != "" {
+		if scaleInMin, err = strconv.ParseInt(v, 10, 64); err != nil {
+			return "", err
+		}
+		if scaleInMax >= 0 {
```
etaoins: Looks like this is a copy-and-paste (should be `scaleInMin`)

scaler/scaler.go
```diff
@@ -9,10 +9,19 @@ import (
 )
 
 type ScaleInParams struct {
```
etaoins: Could these be merged in to `ScaleParams`? They seem symmetric with each other.

lox:
What if rather than min/max or adjustment we went with `ScaleOutFactor` and `ScaleInFactor` (or `Multiplier`), which was a decimal multiplier applied to the `DesiredCount` change? 

That would handle the two main uses cases: 

* Slow down scale in: `ScaleInFactor=0.25` this would reduce scale-ins to only 25% of what they would be otherwise. For instance, scale-in of 10 would end up as a scale in of 3. 

* Faster scale out: `ScaleOutFactor=2` this would increase scale-outs by 200%, so a scale out of 10 would end up 20.

scaler/scaler.go
lox: Yeah, probably. I kind of liked having the type safety and in the current world of positive vs negative changes there didn't seem much economy to combining them.

etaoins:
That seems like it would work better with build queues that have wildly varying load (eg a corporate build queue might have 20 jobs during the day and only a couple overnight). Makes sense to making the scaling scalable 😀

We’d have to make sure we wouldn’t hit an asymptote when scaling down. For example, if we’re scaling down by 25% with only 3 agents we’d need to make sure we eventually scale to 2 instead of always rounding to 3.

lox:
> We’d have to make sure we wouldn’t hit an asymptote when scaling down. For example, if we’re scaling down by 25% with only 3 agents we’d need to make sure we eventually scale to 2 instead of always rounding to 3.

Rounding down should do it, right?

etaoins:
Yup, I think rounding down when scaling down and rounding up when scaling up would handle it.

lox:
Closing in favour of https://github.com/buildkite/buildkite-agent-scaler/pull/12.

PR #13, opened 2019-05-05
Add a configurable scale factor to scale in/out

lox:
This adds a scale-in and scale-out factor that can be used to speed up or slow down scale outs and ins. 

For instance, setting `SCALE_IN_FACTOR=0.5` will reduce the change applied in each scale in by 50%.  Setting `SCALE_OUT_FACTOR=2.0` will increase each scale out change by 200%.

This is a simpler take on #12.

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"strconv"
 	"time"
@@ -16,11 +17,11 @@ import (
 	"github.com/buildkite/buildkite-agent-scaler/version"
 )
 
-var invokeCount = 0
-
-// lastScaleDownTime stores the last time we scaled down the ASG
-// On a cold start this will be reset to Jan 1st, 1970
-var lastScaleInTime time.Time
+// Stores the last time we scaled in/out in global lambda state
+// On a cold start this will be reset to a zero value
+var (
+	lastScaleIn, lastScaleOut time.Time
+)
 
 func main() {
 	if os.Getenv(`DEBUG`) != "" {
@@ -29,8 +30,6 @@ func main() {
 			log.Fatal(err)
 		}
 	} else {
-		invokeCount = invokeCount + 1
-		log.Printf("Invocation count %d", invokeCount)
 		lambda.Start(Handler)
 	}
 }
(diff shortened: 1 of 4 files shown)


lox:
@etaoins would love to hear your thoughts!

lambda/main.go
```diff
@@ -28,54 +29,64 @@ func main() {
...
 
-	if scaleInAdjustmentStr := os.Getenv(`SCALE_IN_ADJUSTMENT`); scaleInAdjustmentStr != "" {
-		var err error
-		scaleInAdjustment, err = strconv.ParseInt(scaleInAdjustmentStr, 10, 64)
-		if err != nil {
+	if v := os.Getenv(`SCALE_IN_FACTOR`); v != "" {
+		if scaleInFactor, err = strconv.ParseFloat(v, 64); err != nil {
 			return "", err
 		}
+		scaleInFactor = math.Abs(scaleInFactor)
```
etaoins: Nice usability tweak :+1:

scaler/scaler.go
```diff
@@ -106,51 +109,136 @@ func (s *Scaler) Run() error {
...
 
-		minimumDesired := current.DesiredCount + s.scaleInParams.Adjustment
-		if desired < minimumDesired {
-			desired = minimumDesired
+	// Calculate the change in the desired count, will be negative
+	change := desired - current.DesiredCount
+
+	// Apply scaling factor if one is given
+	if factor := s.scaleInParams.Factor; factor != 0 {
+		// Use Floor to avoid never reaching upper bound
```
etaoins:
```suggestion
		// Use Floor to avoid never reaching lower bound
```

scaler/scaler.go
```diff
@@ -106,51 +109,136 @@ func (s *Scaler) Run() error {
...
+		factoredChange := int64(math.Floor(float64(change) * factor))
+
+		if factoredChange < change {
+			log.Printf("👮‍️ Increasing scale-in of %d by factor of %0.2f",
```
etaoins: I'm not sure about the utility of having a scale-in factor >100% but if someone is crazy enough to configure that I guess we should honour it.

lox: Yeah, I thought the same 🤷🏼‍♂️

PR #16, opened 2019-05-05
Add detailed readme

lox:
I wrote up a longer README that provides some context.


lox:
@toolmantim I merged this one, but would love feedback on copy!

toolmantim:
Readme is great!

What about letting people know how to run the tests? I was going to suggest using the `bk` command, but that'd run the build and upload stuff too.

lox:
The tests are just the standard golang `go test -v ./...`!

PR #19, opened 2019-05-14
Correct required environment variables in README

mikenicholson:
The MIN_SIZE and MAX_SIZE variables are not used by the lambda. These
parameters are queried from the Autoscaling Group rather than the
environment.


lox:
Thanks @mikenicholson

PR #28, opened 2020-05-11
Update for Serverless Application Repository hosting

keithduncan:
This replaces the per-region S3 buckets with a Serverless Application Repository entry.

The SAR is in us-east-1 but allows you to reference the application from any region. This removes a step when adding support for a new region to the elastic stack. (We still have to copy the AMI(s) to the new region but at least not this too).

I also took this change as an opportunity to _require_ an SSM parameter for the agent token. Inside this serverless application we can strongly type the `BuildkiteAgentTokenParameter` parameter as an SSM parameter that exists and handle the difference between a literal token and SSM parameter name in the enclosing template outside the application, simplifying the code.

So far this is a straight copy of the resources from the elastic-ci stack, I haven’t fiddled with the parameter naming for autoscaling yet, I also haven’t added `Description`s to the parameters.

Currently it supports overriding the lambda IAM role with your own one, but I think it makes more sense to bring in the KMS key ID and completely generate the correct permissions inside the application. I want to experiment with exporting the role ARN from this template and attaching the KMS permission from outside it first.

cc @lox @pda @chloeruka


Makefile
```diff
@@ -21,6 +21,12 @@ endif
 
 build: handler.zip
 
+package: build
+	sam package \
+		--s3-bucket buildkite-sar-us-east-1 \
```
keithduncan, line 26: This is just a bucket in my test account, we’d set up proper CI/CD for the template and SAR resource before merging this

template.yaml
```diff
@@ -64,74 +69,65 @@ Resources:
...
-    Type: AWS::Lambda::Function
-    Condition: UseLambdaAutoscaling
+    Type: AWS::Serverless::Function
     Properties:
-      Code:
-        S3Bucket: !Ref AutoscalingLambdaBucket
-        S3Key: "ec2-agent-scaler.zip"
-        S3ObjectVersion: !Ref AutoscalingLambdaObjectVersion
-      Role: !GetAtt AutoscalingLambdaExecutionRole.Arn
+      CodeUri: handler.zip
```
keithduncan, line 101: `sam package` updates this to point to the s3 URL that the code is uploaded to.

keithduncan:
@yob while I’m working on the agent scaler I’d like to push this over the line so we can adopt it in the elastic stack later, going to look at fixing up the publishing machinery in Buildkite accounts now

yob:
neat!

So this will start publishing new versions to SAR, and we can follow up with a change in the elastic stack that starts using the SAR version instead of the S3 version?

>  also took this change as an opportunity to require an SSM parameter for the agent token. Inside this serverless application we can strongly type the BuildkiteAgentTokenParameter parameter as an SSM parameter that exists and handle the difference between a literal token and SSM parameter name in the enclosing template outside the application, simplifying the code.

Does this mean we'll need to remove the BuildkiteToken parameter from the elastic stack and cut a new major version 6.x?

PR #31, opened 2020-08-17
embrace go modules; delete vendor dir

pda:
This pull request deletes the `vendor/` directory which holds a copy of all dependencies. This existed alongside `go.mod` to support the transition to Go modules. That transition is mature enough now that we can assume we're being built by a version of Go that support modules.

I also ran `go mod tidy` which removed a couple of old checksums and added a `go 1.15` statement to `go.mod` (note: that's not a hard dependency on go >= 1.15).


pda:
> That transition is mature enough now that we can assume we're being built by a version of Go that support modules.

Ironically, our build failed because it was using Go 1.12 🤦🏼‍♂️

#32 fixes that — this PR should be rebased once that merges.

yob:
:+1:

PR #42, opened 2021-05-05
Publish Agent Metrics to Cloudwatch along with existing Job Metrics

theonlysinjin:
This includes the Agent Metrics and sends it along to Cloudwatch.
The metrics were already being collected and have now just been added to the logging and publishing functions

The metrics added:
- IdleAgentCount
- BusyAgentCount
- TotalAgentCount

The main reason to add these is it is possible to scale up the agents when there are none free. This helps us reduce deployment time by having a warm agent running and waiting for the new job (Scale when no free agents), while still getting the useful downscale on idle events.

I have tested that the metrics are there, but I can't get the aws-vault to connect to the ASG, etc to test metrics collection
`2021/05/05 12:43:42 Collecting agent metrics for queue "default"`
`2021/05/05 12:43:43 ↳ Got scheduled=0, running=0, waiting=0, idleAgents=3, busyAgents=0, totalAgents=3 (took 1.102213142s)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,16 @@ See the [forum post](https://forum.buildkite.community/t/experimental-lambda-bas
 
 ## Publishing Cloudwatch Metrics
 
-The scaler collects it's own metrics and doesn't require the [buildkite-agent-metrics][]. It supports optionally publishing the metrics it collects back to Cloudwatch, although it only supports a subset of the metrics that the [buildkite-agent-metrics][] binary collects:
+The scaler collects it's own metrics and doesn't require the [buildkite-agent-metrics][].  
+It supports optionally publishing the metrics it collects back to Cloudwatch.  
 
+Binary collects:
 * Buildkite > (Org, Queue) > `ScheduledJobsCount`
-* Buildkite > (Org, Queue) > `RunningJobCount`
+* Buildkite > (Org, Queue) > `RunningJobsCount`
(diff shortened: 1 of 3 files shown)


keithduncan:
Hi @theonlysinjin thank you for contributing this pull request!

I’ve been taking a look and wondered if you could share more details about your use case? You mention scaling up when there are no idle agents as a desirable behaviour. If the idle count scale out was implemented as an auto scaling policy, this may conflict and fight with the desired capacity set by this lambda.

Setting `scale-out-factor` to a value `>1` might approximate the behaviour you’re looking for by pre-emptively booting more instances than are strictly required at a given instant. Let me know if you think this could work for you, and if not we can try to find the best way to support what you’re looking for. Thanks!

keithduncan:
Hi @theonlysinjin, thank you for opening this pull request. We have been discussing this use case and have chosen to renew support for the [buildkite-agent-metrics Lambda](https://github.com/buildkite/buildkite-agent-metrics) instead of building more metrics in to the scaling Lambda.

These metrics are are already collected and published from the metrics Lambda and it should support your use case.

I also wanted to re-iterate that separately adding an auto scaling policy to the Elastic CI Stack auto scale group may conflict and fight with the desired capacity set by the scaling Lambda, and to exercise caution when mixing these together.

PR #44, opened 2021-05-21
Deal with race condition in pending instances

dbaggerman:
My (untested) hypothesis for https://github.com/buildkite/elastic-ci-stack-for-aws/issues/847 is that there may be a race condition occurring between increasing DesiredInstances and having them be visible in the instance list. This accounts for that by checking if the size of the instances list is less than DesiredInstances. If so, we treat the difference as additional pending
instances.

This is just an educated guess - I haven't replicated the reported problem, so I can't be sure if this is actually the cause or the correct fix.

### DIFF
--- a/scaler/asg.go
+++ b/scaler/asg.go
@@ -42,9 +42,18 @@ func (a *asgDriver) Describe() (AutoscaleGroupDetails, error) {
 		}
 	}
 
+	// There may be a race condition between increasing DesiredCapacity and
+	// actually seeing a pending instance in the list. In this case, treat the
+	// difference as pending instances that are yet to be created.
+	numInstances := int64(len(asg.Instances))
+	desired := int64(*result.AutoScalingGroups[0].DesiredCapacity)
+	if numInstances < desired {
+		pending += desired - numInstances
+	}
+
 	return AutoscaleGroupDetails{
 		Pending:      pending,
-		DesiredCount: int64(*result.AutoScalingGroups[0].DesiredCapacity),
+		DesiredCount: desired,
 		MinSize:      int64(*result.AutoScalingGroups[0].MinSize),
 		MaxSize:      int64(*result.AutoScalingGroups[0].MaxSize),
 	}, nil


yob:
Thanks for thinking about https://github.com/buildkite/elastic-ci-stack-for-aws/issues/847 :heart: 

I'm a bit nervous about the risk of whack-a-mole race conditions around the mixing of our metrics and estimating the ASG size. The root problems we were trying to address in #40 was a "stall" when there's a single instance and it's running fewer agents than AgentsPerInstance.

Just thinking out loud: could we remove the ASG API calls that estimate the number of pending instances, and instead have the scaler decide (solely with our metrics) to bump DesiredSize from 1 to 2 when agentCount < AgentsPerInstance for N minutes?

The awkward thing about that is the scaler would need to maintain state across multiple loops.

dbaggerman:
> I'm a bit nervous about the risk of whack-a-mole race conditions around the mixing of our metrics and estimating the ASG size.

This is the only obvious race condition that I can think of. It was something I thought about while writing the code, but wasn't sure how to test/confirm it - and ended up losing track of it before putting the PR up. It's possible there are others that I haven't anticipated, but I'm not expecting to see much, if any, whack-a-mole.

> could we remove the ASG API calls that estimate the number of pending instances

I didn't actually add any additional API calls for the pending instances. The code already made a `DescribeAutoScalingGroups` API call (to get the current `DesiredCount`), which includes the instance list - we just weren't making use of instance data previously.

> The awkward thing about that is the scaler would need to maintain state across multiple loops.

If you want to maintain state across multiple lambda invocations, then we would have to start persisting state to something like DynamoDB. That seems like quite a bit of additional complexity to avoid a problem which hasn't been confirmed yet. My suggestion would be to start with the simple fix[1] and come back around to a bigger, more comprehensive, refactor/redesign if more edge cases come up.

[1] After we confirm this is actually the problem, and fixes it the way I expect it to.

yob:
I'm seeking some feedback from colleagues and hope to have a proper review soon :heart: 

> I didn't actually add any additional API calls for the pending instances

Sorry, my memory was wrong and I should've checked the code first.

> we would have to start persisting state to something like DynamoDB. That seems like quite a bit of additional complexity to avoid a problem which hasn't been confirmed yet

I am 100% onboard with not adding persistent storage.

> After we confirm this is actually the problem, and fixes it the way I expect it to

Do you have any suggestions for testing this fix prior to cutting an official elastic stack release?

jessebye:
@yob any updates here? We are getting bit by this bug pretty often, which causes long build times plus a spike in cost (since we max out at 50 nodes, and it waits to spin up all 50 before the build starts).

yob:
> which causes long build times plus a spike in cost (since we max out at 50 nodes, and it waits to spin up all 50 before the build starts).

Could you describe these symptoms in some more detail, as they don't align with my expected impact of this issue. Maybe my understanding is wrong, or maybe we have two issues at play.

In particular: no build should wait for all agents to come online, each job in a build can start as soon as a free agent is available. 

Out of interest, what value does your stack use in AgentsPerInstance and do you only see the over-scaling happening when you r connected agents < AgentsPerInstance?

jessebye:
@yob It could be that it just appears to be spinning up all 50 nodes before the build starts, but I don't have any evidence to directly support that. We definitely did have builds waiting ~15 minutes before an agent started work, and the Agents tab in Buildkite was showing 50 agents all green/available. Not sure what would cause that, but it seemed like the newly scaled up agents weren't taking jobs right away. When the bad scale-up doesn't happen, everything works as expected - a new agent comes online, and the job is picked up pretty quickly.

Our stack is using 2 AgentsPerInstance. The overscaling _does_ seem to only happen when we have connected agents < AgentsPerInstance.

dbaggerman:
Thinking this through, there are a few state changes between changing `DesiredCount` and getting an agent registered. We change `DesiredCount`, then a `pending` instance is created, then once the instance is provisioned the agent needs to register to Buildkite.

The original code watched for the `pending` instance to instance provisioned stage, and this PR would cover a potential gap[1] between `DesiredCount` changing and a `pending` instance being created.

What @jessebye is describing sounds like it could be a gap between the instance being provisioned (as far as the ASG is concerned) and the agent registering to buildkite. This would result in the lambda seeing fully provisioned instances but no available agents, which would trigger the anti-stall behaviour.

The lambda also uses the internal agent metrics API to check how many agents are registered, rather than the public agents API. @yob, is there maybe some caching on that API which might cause a delay between an agent registering and the new total being available to the agent metrics?

But if @jessebye is seeing agents registered, but not accepting jobs, that sounds like a different (but possibly related) problem.

[1] I would think in most cases this gap is small enough that it'll pass by the time the lambda does its next poll. If so, this patch may not be strictly necessary but it wouldn't hurt to cover it anyway.

keithduncan:
Going to close this one as I didn’t observe this (not to say it isn’t happening 🔍) but did observe a significant gap between an instance moving out of pending before there are any agents registered. Once we handle that larger case better we can take another look at whether there are any finer grained races in the ASG control pane.

PR #46, opened 2021-06-08
Revert agent shortfall scaling

keithduncan:
This was added in #40 but is causing issues where the stack will scale too aggressively.

There is a race condition between an instance being added to the desired count in the ASG and the buildkite agents being registered while the cloud-init bring up is completed.

One way to handle this might be to add a boot time lifecycle hook, to hold the instance in the ASG `Pending` state until the Buildkite agents are booted and can be reliably counted. However, in the short term I think we need to revert this behaviour and roll out an updated stack for customers incurring undesirable additional cost.

I’ve left the additional refactoring in place from that pull request and have just removed the change in the calculation.

### DIFF
--- a/scaler/scaling_calculator.go
+++ b/scaler/scaling_calculator.go
@@ -33,25 +33,5 @@ func (sc *ScalingCalculator) DesiredCount(metrics *buildkite.AgentMetrics, asg *
 		desired = sc.perInstance(agentsRequired)
 	}
 
-	// If there are less agents registered than we'd expect based on the size
-	// of the autoscaling group, then there may be instances not accepting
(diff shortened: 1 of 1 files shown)


yob (approved):
I think this is the best course of action for now, sorry @dbaggerman :pray: 

We're chatting about alternative approaches to avoiding the stalling issue, but in the meantime avoiding the overscaling (and increased cost) issues experienced by other customers is a priority.

dbaggerman:
That's understandable @yob. Since we're not using the upstream `elastic-ci-stack`, we can keep our own stack pinned to 1.1.0 for now - as long as it doesn't take too long to come up with an alternative, rolling this back won't cause us any major pain.

PR #48, opened 2021-07-13
Use AWS shared credentials config

lox:
This enables shared config checks in the AWS Session SDK for setups that use `~/.aws/credentials`. See https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html

Note that 508764a makes a change where by a single session is used and passed around vs `session.New()` invoked everywhere. It's remotely possible this could break things where creating a new session each time was providing some durability around failures and dropouts unintentionally. I don't think this is likely, as there are internal retry mechanisms in the single session. In fact, I think that it's likely that establishing lots of new sessions could lead to rate-limiting and latency.

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-lambda-go/lambda"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/buildkite/buildkite-agent-scaler/buildkite"
 	"github.com/buildkite/buildkite-agent-scaler/scaler"
 	"github.com/buildkite/buildkite-agent-scaler/version"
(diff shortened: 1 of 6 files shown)


keithduncan (approved):
I wonder whether we should just have this on all the time for the local `main.go` as the SDK still preferences the environment variables. We’d still use the environment variables from something like `aws-vault` even if you have a `~/.aws/config`.

Looks good either way to me!

keithduncan:
Realised after approving that you can’t hit the merge button here, let me know what you think about making this always-on vs a flag then I can merge or we’ll tweak it and merge 🙇‍♂️

lox:
Yeah, after some thought just as a default makes sense.

lox:
I'd suggest merge and squash for this one to remove the confusing history!

keithduncan (approved):

PR #49, opened 2021-08-03
Add support for Agent Tokens stored in Secrets Manager

lox:
This adds support for accessing Agent Tokens directly in AWS Secrets Manager.

It's arguable that one could just access the secret directly via `/aws/reference/secretsmanager/${SECRET}`. (See https://docs.amazonaws.cn/en_us/systems-manager/latest/userguide/integration-ps-secretsmanager.html).

Presented for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ An AWS Lambda bundle is created and published as part of the build process. The 
 
 It's entrypoint is `handler`, it requires a `go1.x` environment and requires the following env vars:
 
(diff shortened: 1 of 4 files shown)


keithduncan:
Thanks for pushing this up @lox!

> It's arguable that one could just access the secret directly via `/aws/reference/secretsmanager/${SECRET}`.

Do you think there are benefits of direct Secrets Manager support over the SSM reference syntax?

lox:
After some thought, I don't think this adds anything. With documentation on how to use the SSM reference syntax it would be fine as is.

PR #51, opened 2021-10-04
Scale only after cooldown plus scale in and out event

gu-kevin:
Currently, scale in occur after cool down period plus last scale in event. However, if there was a scale in event a day ago and then there was a a scale out event, it is possible for it to immediately scale in because the cool down period is only relative to the last scale in event.

To scale in and out based on the last event + cool down period, we can set Set SCALE_ONLY_AFTER_ALL_EVENT environment to scale in and out after the last event + cooldown period corresponding to the same type of event.


gu-kevin:
got it that makes sense

PR #52, opened 2021-10-04
Set last scale in and out event on autoscaler activity

gu-kevin:
Currently, the scale in and out event time is set and stored as a global variable. However, if the service does not run in the same container, the event is always reset to 0 and will scale in and out immediately. This change will set the last scale in and out event based on the autoscaler's activity event, so if the service does not run on the same lambda container, it will keep the state of the last time it scaled in or out.

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -18,6 +18,12 @@ import (
 	"github.com/buildkite/buildkite-agent-scaler/version"
 )
 
+// Stores the last time we scaled in/out in global lambda state
+// On a cold start this will be reset to a zero value
+var (
+	lastScaleIn, lastScaleOut time.Time
+)
+
 func main() {
 	if os.Getenv(`DEBUG`) != "" {
 		_, err := Handler(context.Background(), json.RawMessage([]byte{}))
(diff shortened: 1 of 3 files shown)


gu-kevin:
I think relying on the API is more reliable on container because that state can change anytime.

the data of when the last scaling events are only in autoscaler activity and the api does not provide any information besides strings that you can filter through.
Screenshot 2021-09-29 10 43 30 AM

keithduncan:
> I am referring to the overall time for the lambda function.

Okay great, that’s the expected execution duration for the lambda function 😄 

What do you think about adding some timestamped logs around `asg.GetLastScalingInAndOutActivity()` so that we have a record of how long that aspect of the function is taking to run?

I think defaulting the ASG scale activity discovery duration to 10s (`asgActivityTimeoutDuration` in your code) before the lambda would fail forward *without* restoring the scale activities, would be reasonable. Adding some timestamped logs before and after so we have a record of how long this is taking in practice, and logging both end cases (success or any of the failure outcomes) will also let us see how this is behaving in practice.

PR #59, opened 2021-11-22
Parameterise DisableScaleIn in the template

keithduncan:
Adds a new optional `DisableScaleIn` parameter that defaults to `true`, the current value. This allows importing stacks to set this to `false` and enable the scale in behaviour of the scaling lambda.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -52,6 +52,14 @@ Parameters:
       - "false"
     Default: "false"
 
+  DisableScaleIn:
+    Description: Whether the desired count should ever be decreased on the given Auto Scaling group. Defaults to true, instances are expected to self terminate when they are idle.
+    Type: String
+    AllowedValues:
+      - "true"
+      - "false"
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -150,7 +158,7 @@ Resources:
           BUILDKITE_QUEUE:               !Ref BuildkiteQueue
           AGENTS_PER_INSTANCE:           !Ref AgentsPerInstance
           CLOUDWATCH_METRICS:            "1"
-          DISABLE_SCALE_IN:              "1"
+          DISABLE_SCALE_IN:              !Ref DisableScaleIn
```
keithduncan, line 161: While this changes the value passed from `"1"` to `"true"` the lambda code already [supports](https://github.com/buildkite/buildkite-agent-scaler/blob/8f7d8e2caca3808a75a33164a0720a5dae479286/lambda/main.go#L176) this.

PR #70, opened 2022-08-10
[FDN-832] Add Description property to Lambda Function

ellsclytn:
Useful from a billing & resource management perspective, as it can
provide a little bit of extra context on what the Function is used for.


triarius (approved):
LGTM

PR #72, opened 2022-10-13
Allow running scaler with a fixed size instance buffer

wbond:
This will help fleets where instance startup time is slow, by allowing them to configure enough free instances to handle a certain amount of load while new instances spin up.


wbond:
Hi @moskyb - wondering if I could get a review on this to see if it is worth continuing to pursue?

jpfielding:
i was literally just coming here to look for this, bump

PR #75, opened 2022-12-15
Add parameter to configure logs retention

Amir-Ahmad:
This PR adds a Cloudformation parameter to configure the cloudwatch log retention. This is currently hardcoded to 1 day, and we'd like to set this to a higher number like 180 days.


triarius (approved):
Thanks for the contribution @Amir-Ahmad this looks good to me

PR #88, opened 2023-05-15
Replace all unnecessary backtick strings with reggo strings

moskyb:
[super nit]


triarius (approved):
💜

triarius:
The conflict should be pretty easy to resolve

PR #93, opened 2023-07-14
Allow the event schedule rate to be configured via parameters

tomellis91:
## Description
This PR is being raised to allow the Event Schedule rate to be configurable via Parameter inputs. 
There is a Default value provided to ensure the change is backwards compatible. 

## Reason for the change
When there are a large number of BuildKite agent queues combined with other resources such as Cluster Autoscalers the AWS API for `DescribeAutoScalingGroupsInfo` receives a `ThrottlingException`. This has a negative impact on all services that are required to interact with the AWS Autoscaling API. 

This PR will allow for the schedule to be updated as required to ensure throttling exceptions are not triggered. 

If this PR is merged, another PR will need to be raised against the elastic CI stack to allow the parameter to be passed in.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -33,6 +33,11 @@ Parameters:
     Type: String
     Default: ''
 
+  EventScheduleRate:
+    Description: How often the Event Schedule is triggered (in minutes)
+    Type: String
+    Default: "1"
+
   AgentsPerInstance:
     Description: ""
     Type: Number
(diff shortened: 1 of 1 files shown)


tomellis91:
hey @triarius , 
Apologies for tagging you directly. I don't seem to have the ability to assign the PR. 
We are currently getting throttled and our CAs are being impacted. Could you please review this if possible and see if it is something you would be happy to merge in?

Cheers

PR #94, opened 2023-07-25
Add MinPollInterval param

### LINKED ISSUE #71: Allow Sleep interval to be configurable
Currently, we are locked into a 10 second sleep period between runs or less if minPollDuration is less than 10 seconds. This should change so we can pass in any reasonable value say from 0-60 with the default being 10.

DrJosh9000:
This controls the sleep between polls. While it could be set to almost any duration, anything smaller than 10s is overridden by Buildkite-Agent-Metrics-Poll-Duration. 

Fixes #71


triarius (approved):
LGTM 🎉

PR #95, opened 2023-07-25
DescribeScalingActivities once

DrJosh9000:
And use `context.WithTimeout` technology instead of separate-goroutine-channel-wait technology.

And move the client and scaler creation outside the loop.

Plus minor cleanups with `mustGetEnv{,Int}`.


triarius (approved):
Thanks for making sense of this. LGTM.

PR #97, opened 2023-08-14
Create a new release process

triarius:
Previously, we conditionally uploaded steps that created a release on GitHub and uploaded the lambda to s3. Now, I've made those steps be conditional steps instead.

I've also moved the block step to be after the release on GitHub is created. The idea is that the release is just a draft. Once the draft is acceptable, the pipeline should be unblocked, pushing the lambda to s3. The same should be done for the SAR release pipeline, which needs different permissions, so lives in a different Buildkite organization. This second pipeline repeats some work from the main pipeline, so I would like to remove it eventually.

I have also changed the release step to read the CHANGELOG.md file and use the GitHub CLI instead of a custom tool use used to develop.

The new procedure is here: https://github.com/buildkite/buildkite-agent-scaler/blob/pdp-1449-makebuildkite-agent-scaler-be-published-to-s3-again/RELEASE.md


triarius:
I'm going to merge this and iterate on the default branch to get the release process working again.

PR #102, opened 2023-08-28
Change `EventScheduleRate` parameter to `EventSchedulePeriod` and require units

triarius:
The rate expression must pluralise the unit, so `1 minute` and `0 minutes` are valid, but `0 minute` is not.

Rather than condition on the value of the parameter, we leave this to
the user to specify whether they should use "minute" or "minutes". This
also allows them to use other units like 1 hour, 1 day, etc. Though
units that are longer or shorter than a minute are not recommended.

Since the old parameter was essentially broken (it would not succeed with any value other than the default, `1`), we've decided to rename the parameter to `EventSchedulePeriod`. This name is more accurate - a length of time like 1 minute is not a rate.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -33,10 +33,10 @@ Parameters:
     Type: String
     Default: ''
 
-  EventScheduleRate:
-    Description: How often the Event Schedule is triggered (in minutes)
+  EventSchedulePeriod:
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
🙌

PR #105, opened 2023-10-05
Add dependabot

triarius:
Some of the dependencies are ancient. Adding Dependabot now. I may tweak some options later.


DrJosh9000 (approved):
S(h)ip it 🍸

PR #139, opened 2023-11-23
Update README example due to end of support for the Go 1.x runtime

tomowatt:
README example needs updated to show new runtime to be used rather than one to be made deprecated


triarius (approved):
Thanks! Nice catch!

PR #179, opened 2024-03-01
Calm dependabot down a little

yob:
We added dependabot in #105 (Oct 2023) and it got very excited proposing daily updates to github.com/aws/aws-sdk-go. 

Eventually it gave up.

Maybe monthly is enough to stop dependencies falling behind, without quite so much noise?


triarius (approved):
Thanks! I used to respond to these, but they became too spammy after a while.

PR #188, opened 2024-07-17
Scaler cleanups

DrJosh9000:
This is a grab-bag of code cleanups.

The only significant change should be that `scaler.Run` now happens before checking `timeout`, so the lambda won't time out before it does anything useful (regardless of `LAMBDA_TIMEOUT`).

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -81,69 +81,75 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
 	)
 
 	if v := os.Getenv("LAMBDA_INTERVAL"); v != "" {
-		if interval, err = time.ParseDuration(v); err != nil {
+		d, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
 		}
+		interval = d
 	}
 
 	if v := os.Getenv("LAMBDA_TIMEOUT"); v != "" {
-		if timeoutDuration, err := time.ParseDuration(v); err != nil {
+		d, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
-		} else {
-			timeout = time.After(timeoutDuration)
 		}
+		timeout = time.After(d)
 	}
 
 	if v := os.Getenv("ASG_ACTIVITY_TIMEOUT"); v != "" {
-		if timeoutDuration, err := time.ParseDuration(v); err != nil {
+		d, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
-		} else {
-			asgActivityTimeoutDuration = timeoutDuration
 		}
+		asgActivityTimeoutDuration = d
 	}
 
 	if v := os.Getenv("SCALE_IN_COOLDOWN_PERIOD"); v != "" {
-		if scaleInCooldownPeriod, err = time.ParseDuration(v); err != nil {
+		p, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
 		}
+		scaleInCooldownPeriod = p
 	}
 
 	if v := os.Getenv("SCALE_IN_FACTOR"); v != "" {
-		if scaleInFactor, err = strconv.ParseFloat(v, 64); err != nil {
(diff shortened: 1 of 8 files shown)


lambda/main.go
```diff
@@ -80,70 +65,97 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
...
+	}
+
+	agentsPerInstanceStr := os.Getenv("AGENTS_PER_INSTANCE")
+	if agentsPerInstanceStr == "" {
+		return "", errors.New("AGENTS_PER_INSTANCE is required")
+	}
+	agentsPerInstance, err := strconv.Atoi(agentsPerInstanceStr)
+	if err != nil {
+		return "", fmt.Errorf("AGENTS_PER_INSTANCE must be an integer: %w", err)
+	}
```
wolfeidau, line 86: I would prefer a function for this

lambda/main.go
```diff
@@ -80,70 +65,97 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
...
+
 	if v := os.Getenv("LAMBDA_INTERVAL"); v != "" {
-		if interval, err = time.ParseDuration(v); err != nil {
+		d, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
 		}
+		interval = d
```
wolfeidau, line 93: Should you just assign this and remove the var?

lambda/main.go
```diff
@@ -80,70 +65,97 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
...
+		asgActivityTimeoutDuration = d
 	}
 
 	if v := os.Getenv("SCALE_IN_COOLDOWN_PERIOD"); v != "" {
-		if scaleInCooldownPeriod, err = time.ParseDuration(v); err != nil {
+		p, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
 		}
+		scaleInCooldownPeriod = p
```
wolfeidau, line 118: As above, should you just assign this and remove the var?

lambda/main.go
```diff
@@ -80,70 +65,97 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
...
 
 	if v := os.Getenv("ASG_ACTIVITY_TIMEOUT"); v != "" {
-		if timeoutDuration, err := time.ParseDuration(v); err != nil {
+		d, err := time.ParseDuration(v)
+		if err != nil {
 			return "", err
-		} else {
-			asgActivityTimeoutDuration = timeoutDuration
 		}
+		asgActivityTimeoutDuration = d
```
wolfeidau, line 110: As above, should you just assign this and remove the var?

lambda/main.go
```diff
@@ -80,70 +65,97 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
...
 		}
-		scaleInFactor = math.Abs(scaleInFactor)
+		scaleInFactor = math.Abs(f)
 	}
 
 	if v := os.Getenv("SCALE_ONLY_AFTER_ALL_EVENT"); v != "" {
-		if v == "true" || v == "1" {
-			scaleOnlyAfterAllEvent = true
-		}
+		scaleOnlyAfterAllEvent = v == "true" || v == "1"
```
wolfeidau, line 130: This seems like it could be a bad idea, maybe a function?

wolfeidau (changes requested):
Minor observations and nitpicks.

lambda/main.go
DrJosh9000: I've refactored the environment variable loading back towards a few helper functions, which I feel also resolves the comments below. Since they all error out of the lambda handler, making it useless, I've leaned towards `log.Fatalf`.

wolfeidau (approved):

lambda/main.go
wolfeidau: 👍🏻

PR #200, opened 2024-12-09
Update the role used to publish SAR

yob:
The new roel has the same permissions, only the name has changed


DrJosh9000 (approved):

yob:
We don't have a release brewing so I won't test this right now. When we get to a release, if anything doesn't work as expected please flail in my direction and I can help sort it out ❤️

PR #203, opened 2025-01-20
Add an option to override the agent endpoint URL via flag and env var

wolfeidau:
I have this new endpoint in a way that always defaults to https://agent.buildkite.com/v3 if not provided to ensure compability with old installations or forks.

I used `AgentEndpoint` for the param name in the template as it matches the name in the elastic stack template.


DrJosh9000 (approved):
LGTM ✅

PR #221, opened 2025-06-12
fix: scope down ElasticCIModeSSMAndEC2 policy

scadu:
As per [https://docs.aws.amazon.com/systems-manager/latest/userguide/run-command-setting-up.html#tag-based-access](https://docs.aws.amazon.com/systems-manager/latest/userguide/run-command-setting-up.html#tag-based-access), we need two statements to restrict SSM command runs.

1. `ssm:SendCommand` and `ssm:GetCommandInvocation` are now conditional on target EC2 instances possessing **both**:    
    - `Role: buildkite-agent` tag
    - `aws:autoscaling:groupName` tag matching the `!Ref AgentAutoScaleGroup` parameter passed down from the Elastic CI Stack CloudFormation Stack
2. `ec2:TerminateInstances` is now conditional on target EC2 instances possessing **both**:
    - `Role: buildkite-agent`
    - `aws:autoscaling:groupName` tag matching the `!Ref AgentAutoScaleGroup` parameter passed down from the Elastic CI Stack CloudFormation Stack

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -250,15 +250,17 @@ Resources:
                   Action:
                     - ssm:SendCommand
                     - ssm:GetCommandInvocation
-                  Resource: '*'
+                  Resource:
+                    - !Sub "arn:aws:ssm:${AWS::Region}::document/AWS-RunShellScript"
                 - Effect: Allow
                   Action:
                     - ssm:SendCommand
                     - ssm:GetCommandInvocation
                   Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
                   Condition:
-                    StringLike:
-                      "ec2:ResourceTag/aws:autoscaling:groupName": !Sub "${AWS::StackName}-AgentAutoScaleGroup-*"
+                    StringEquals:
+                      "ssm:resourceTag/Role": "buildkite-agent"
+                      "ssm:resourceTag/aws:autoscaling:groupName": !Ref AgentAutoScaleGroup
                 - Effect: Allow
                   Action:
                     - ssm:DescribeInstanceInformation
@@ -270,8 +272,9 @@ Resources:
                     - ec2:TerminateInstances
                   Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
                   Condition:
-                    StringLike:
-                      "ec2:ResourceTag/aws:autoscaling:groupName": !Sub "${AWS::StackName}-AgentAutoScaleGroup-*"
+                    StringEquals:
+                      "ec2:ResourceTag/Role": "buildkite-agent"
+                      "ec2:ResourceTag/aws:autoscaling:groupName": !Ref AgentAutoScaleGroup
           - !Ref 'AWS::NoValue'
 
   AutoscalingFunction:
--- a/template.yaml
+++ b/template.yaml
@@ -247,20 +247,19 @@ Resources:
               Version: '2012-10-17'
               Statement:
                 - Effect: Allow
-                  Action:
-                    - ssm:SendCommand
-                    - ssm:GetCommandInvocation
+                  Action: ssm:SendCommand
                   Resource:
                     - !Sub "arn:aws:ssm:${AWS::Region}::document/AWS-RunShellScript"
                 - Effect: Allow
-                  Action:
-                    - ssm:SendCommand
-                    - ssm:GetCommandInvocation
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -250,15 +250,17 @@ Resources:
                   Action:
                     - ssm:SendCommand
                     - ssm:GetCommandInvocation
-                  Resource: '*'
+                  Resource:
+                    - !Sub "arn:aws:ssm:${AWS::Region}::document/AWS-RunShellScript"
```
lucaswilric, line 254: This will work for the `ssm:SendCommand` action. However, it is not possible to limit the `ssm:GetCommandInvocation` action by resource¹.

To get the resource restricction for `ssm:SendCommand`, you separate the 2 actions into different statements.

---

¹ In the [IAM Actions table for SSM](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awssystemsmanager.html), there are no entries in the "Resource types" column for `GetCommandInvocation`. This means we unfortunately have to set `Resource: "*"`.

template.yaml
```diff
@@ -250,15 +250,17 @@ Resources:
...
                   Action:
                     - ssm:SendCommand
                     - ssm:GetCommandInvocation
                   Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
                   Condition:
-                    StringLike:
-                      "ec2:ResourceTag/aws:autoscaling:groupName": !Sub "${AWS::StackName}-AgentAutoScaleGroup-*"
+                    StringEquals:
+                      "ssm:resourceTag/Role": "buildkite-agent"
+                      "ssm:resourceTag/aws:autoscaling:groupName": !Ref AgentAutoScaleGroup
```
lucaswilric, line 263: As above, this `Condition` is OK for `ssm:SendCommand`, but not for `ssm:GetCommandInvocation`. Even though it's annoyingly repetitive, I recommend separating the 2 Actions into distinct Statements.

template.yaml
```diff
@@ -270,8 +272,9 @@ Resources:
                     - ec2:TerminateInstances
                   Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
                   Condition:
-                    StringLike:
-                      "ec2:ResourceTag/aws:autoscaling:groupName": !Sub "${AWS::StackName}-AgentAutoScaleGroup-*"
+                    StringEquals:
+                      "ec2:ResourceTag/Role": "buildkite-agent"
+                      "ec2:ResourceTag/aws:autoscaling:groupName": !Ref AgentAutoScaleGroup
```
lucaswilric, line 277: As with `ssm:GetCommandInvocation`, `ssm:DescribeInstanceInformation` does not support `Resource`s or `Condition`s. Unfortunately we have to grant `*` for that one.

The `Condition`s look fine for EC2 though 👍

lucaswilric (commented):
I think this will need some changes in order to work in all cases. See comments.

Caveat: I'm unsure whether IAM will fail open or closed when it encounters a `Condition` that is incompatible with the resource type. I've assumed it fails closed. In any case, it's better to keep as close to strict compatibility as we can, even if it turns out my assumption is wrong.

template.yaml
scadu: @lucaswilric `- ssm:DescribeInstanceInformation` is already pinned to `Resource: '*'` due to this issue, so I think we are fine here? 
Condition is applied to `ec2:TerminateInstances` Action.

template.yaml
scadu: Missed that one! Addressed now. Thanks, @lucaswilric!

lucaswilric (commented):
This looks good to me, as far as the scope of [this review request](https://linear.app/buildkite/issue/SECT-41/request-to-review-policy-role-for-buildkite-agent-scaler).

That is, I don't see any security issues or IAM-related pitfalls. I have not personally validated the changes against requirements (though I believe @scadu has).

lucaswilric (approved):

scadu:
> This looks good to me, as far as the scope of [this review request](https://linear.app/buildkite/issue/SECT-41/request-to-review-policy-role-for-buildkite-agent-scaler).
> 
> That is, I don't see any security issues or IAM-related pitfalls. I have not personally validated the changes against requirements (though I believe @scadu has).

Thanks for reviewing, @lucaswilric! Tested the template against my own Stack — works as expected!

PR #222, opened 2025-06-17
feat: update tooling to use mise

scadu:
What's mise? Tooling version manager, in short. It does provide additional stuff like a task runner, or environment variable management, but let's focus on tooling management.

### Why?

Currently, we use mix of Homebrew and containers (mostly in CI pipeline itself).Results can be... unpredictable. One can happily run tests, pre-commit checks can be all fine, only to see a failed build because $TOOL is in a different version (e.g. pinned container image vs. newest Homebrew package one have installed on their machine).

### How?

With mise.toml committed to a repository (mise is also compatible with asdf `.tool-versions` files as well as [idiomatic version files](https://mise.jdx.dev/configuration.html#idiomatic-version-files) like `.node-version` and `.ruby-version`. See [configuration](https://mise.jdx.dev/configuration.html) for more details.), mise would ensure same tooling is available locally and in CI.

No need to guess what's needed, no hacks to run $TOOL in $VERSION, no fetching a dozen of containers just to run a linter.

https://mise.jdx.dev/continuous-integration.html might be worth exploring, but in my case, I just include eval `"$(mise activate bash)" && mise install` in `pre-command` hook for the agent.

Okay, but how do I use specific version of a tool in my project? 

Locally, start from installing Mise: `brew install mise`. I'd also recommend adding Mise to your shell config, see https://mise.jdx.dev/getting-started.html#activate-mise.

Let's say we want Go 1.23 in our project, but we also have Go 1.24 installed locally, damn!With mise we can pin Go's version in the project by running: `mise use use [go@1.23](mailto:go@1.23)`.Sample output:

```
mise ~/Development/buildkite-agent-scaler/mise.toml tools: go@1.23.9
mise +go@1.23.9
mise -go@1.22.12
```

Boom, we've now pinned Go to the latest version (at the time of running the command) of the 1.23 release!

In fresh projects, we could just execute `mise use go --pin` to use latest version

___
`bin/mise` has been generated with https://mise.jdx.dev/continuous-integration.html#bootstrapping

### DIFF
--- a/.buildkite/pipeline-sar.yaml
+++ b/.buildkite/pipeline-sar.yaml
@@ -1,15 +1,24 @@
+# YAML anchor for mise setup
+x-mise-setup: &mise-setup |
+  if [[ ! -f ./bin/mise ]]; then
+    mkdir -p ./bin
+    curl -fsSL https://mise.jdx.dev/mise-latest-linux-x64 > ./bin/mise
+    chmod +x ./bin/mise
+  fi
+  ./bin/mise install
+
 steps:
   - name: ":golang: Tests"
     key: test
-    command: .buildkite/steps/tests.sh
-    plugins:
-      - docker#v3.1.0:
-          image: "golang:1.24"
-          workdir: /go/src/github.com/buildkite/buildkite-agent-scaler
+    commands:
+      - *mise-setup
+      - .buildkite/steps/tests.sh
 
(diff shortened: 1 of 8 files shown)


tomowatt:
@omehegan can you remove any links to private documents as this is a public repository and shouldn't be shared here.

PR #246, opened 2025-10-04
add interval as command line args

yuanyeyuanye:
add interval as command line args


petetomasik (commented):
👋 Hi @yuanyeyuanye, thanks for the submission! Is the idea here to allow the `interval` to be configurable above the `10` seconds returned by the `Buildkite-Agent-Metrics-Poll-Duration` response header?

PR #257, opened 2025-11-06
Fix availability-based scaling and simplify threshold logic

scadu:
The availability check detected low agent availability but only acted on
it in Elastic CI Mode. Without Elastic CI Mode, it would log the warning
but not scale up. 
Added else clause to handle non-Elastic CI Mode by
adding +1 instance when availability drops below threshold.

Also simplified the availability threshold condition.


scaler/scaler_test.go
petetomasik: We are missing tests for `availabilityThreshold` that evaluates `actualCount`.

scadu: @petetomasik done!

Issue #269, opened 2025-11-26
Unreasonable instance count error


noahtallen:
Hi, we recently triggered [this error](https://github.com/buildkite/buildkite-agent-scaler/blame/c075984578afef4ba42d71aa6f7301f8838068b4/scaler/scaling_calculator.go#L32) ("calculated unreasonable instance amount xyz, capping at 1000") after updating to a recent version of the agent scaler.

We typically run more than 1000 instances, so this was unexpected -- it seems that when `agentsPerInstance` is set to 0, the function is ok returning the raw number of required instances (1:1 mapping), but if set to 1 (also 1:1 mapping), it caps it at a hardcoded 1000. Is that intentional or do we need to change our configuration after the update?

PR #270, opened 2025-11-26
SUP-5647: Make instance cap configurable via MAX_INSTANCE_CAP

Mykematt:
Previously the instance count was hardcoded to cap at 1000, which caused issues for customers running more than 1000 instances.

This change introduces a new MAX_INSTANCE_CAP environment variable that allows users to configure the maximum instance count cap. Setting it to 0 (the default) means no cap is applied.

Fixes the issue where agentsPerInstance=1 would cap at 1000 while agentsPerInstance=0 (which defaults to 1:1 mapping) had no cap.

Fixes issue #269


Mykematt:
> Looking at the code now, we might remove the cap, since there's already `MIN_SIZE` and `MAX_SIZE`. Although we should test how the proportional buffer behaves after changing that.

That make sense. What would you suggest as a replacement? All lower case as in `max_instance_cap`? I do think `MAX_SIZE` and `MAX_INSTANCE_CAP` are intuitively and operationally different though.

PR #274, opened 2025-12-10
SUP-2919: Add ManagedPolicyARNs to Lambda Scaler

Mykematt:
Adds a new `ManagedPolicyARNs` parameter to the CloudFormation template, allowing users to attach additional managed IAM policies to the Lambda execution role. 

It ensures that the default `AWSLambdaBasicExecutionRole` policy is always included.


Mykematt:
> @Mykematt can we get a description added to this PR please? It would be helpful for those reviewing it

Done.

PR #305, opened 2026-04-22
Handle stale instance IDs in DescribeInstances batch

scadu:
### Problem

`CleanupDanglingInstances` puts every ASG InstanceID into one `ec2:DescribeInstances` call. 
If any one of them has terminated between the ASG Describe and the EC2 Describe, AWS returns `InvalidInstanceID.NotFound` and the whole batch fails.

### Fix
New `describeInstancesTolerant` helper wraps the call. On `InvalidInstanceID.NotFound` (detected via `smithy.APIError`), it pulls the stale IDs out of the message, drops them, retries once. 
If everything was stale, it returns an empty output and the cleanup loop does nothing. Any other error, or a second `NotFound` on the retry, is returned as-is – one retry, no loops.

### DIFF
--- a/scaler/asg.go
+++ b/scaler/asg.go
@@ -451,9 +451,7 @@ func (a *ASGDriver) CleanupDanglingInstances(ctx context.Context, minimumInstanc
 		return nil
 	}
 
-	descInstancesOutput, err := ec2Client.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
-		InstanceIds: asgDetails.InstanceIDs,
-	})
+	descInstancesOutput, err := describeInstancesTolerant(ctx, ec2Client, asgDetails.InstanceIDs, a.Name)
 	if err != nil {
 		return fmt.Errorf("failed to describe instances in ASG %s: %w", a.Name, err)
 	}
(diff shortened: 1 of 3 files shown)


scaler/asg.go
```diff
@@ -448,9 +451,7 @@ func (a *ASGDriver) CleanupDanglingInstances(ctx context.Context, minimumInstanc
 		return nil
 	}
 
-	descInstancesOutput, err := ec2Client.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
-		InstanceIds: asgDetails.InstanceIDs,
-	})
```
petetomasik, line 453: Do we also need to apply the same update to use `describeInstancesTolerant()` in `scaler.go` - https://github.com/buildkite/buildkite-agent-scaler/blob/b57bb2e446314e9c92ce434e58441d7bc46b098f/scaler/scaler.go#L424-L426 ?

scadu: There's already a fallback in the flow you've linked to. If we swapped over to `describeInstancesTolerant()` in the `scaler.go`, I'd do that in a follow up PR as tests would also had to be extended.
That's unless you strongly feel we should do it at once.

petetomasik (approved):

PR #310, opened 2026-05-22
Use adaptive AWS retryer

scadu:
The SDK default (standard, 3 attempts, no client-side rate limiting) lets throttling errors through when many scaler Lambdas hit the same AWS APIs at once. Adaptive mode adds a token bucket that backs off when AWS throttles.

New `scaler.LoadAWSConfig` wraps `config.LoadDefaultConfig` for the CLI and Lambda entry points and skips its defaults if `AWS_RETRY_MODE` or `AWS_MAX_ATTEMPTS` is set.

Refs SUP-6910, SUP-6908.

### DIFF
--- a/lambda/main.go
+++ b/lambda/main.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-lambda-go/lambda"
-	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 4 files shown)


JoeColeman95 (approved):
LGTM, with the adaptive mode being the right call IMO.

PR #312, opened 2026-05-22
Slide the dangling-instance check across the list

scadu:
The scaler picks the oldest N instances and probes each via SSM. If those N are unreachable (the most common zombie symptom), the same N get retried every minute and the rest of the fleet is never examined. 
On a 100-instance ASG with N=5, zombies past position 5 stay invisible.

rotateInstanceWindow advances the slice by N every EventPeriodSec, covering the fleet in ceil(total/N) ticks.
Wall clock time is the seed – Lambda doesn't keep any state.

Marking instances with unreachable SSM agent as dangling is going to be done in a separate PR.


scaler/scaler.go
```diff
@@ -41,6 +41,7 @@ type Params struct {
 	MinimumInstanceUptime       time.Duration // How long instance should be online before being eligible for dangling instance check
 	MaxDanglingInstancesToCheck int           // Maximum number of instances to check for dangling instances (only used for dangling instance scanning, not for normal scale-in)
 	MaxInstanceCap              int           // Maximum instance count cap (0 means no cap)
+	EventPeriodSec              int64         // Tick size in seconds for the dangling-check rotation. Defaults to 60 when 0.
```
petetomasik: Should this be `time.Duration` instead of `int64` ?

lambda/main.go
```diff
@@ -191,6 +191,7 @@ func Handler(ctx context.Context, evt json.RawMessage) (string, error) {
 		MinimumInstanceUptime:       minimumInstanceUptime,
 		MaxDanglingInstancesToCheck: maxDanglingInstancesToCheck,
 		MaxInstanceCap:              maxInstanceCap,
+		EventPeriodSec:              int64(interval.Seconds()),
```
petetomasik: Instead of `EventPeriodSec`, should this instead be named something like `DanglingInstancesCheckInterval` (or similar)?

petetomasik (approved):

PR #314, opened 2026-05-29
Batch dangling-instance SSM check

scadu:
Before, `checkAndMarkUnhealthy` sent one `SendCommand` per instance and
polled `GetCommandInvocation` in a serial loop, so it got slower as the instance count grew.

This PR changes that to:

1. `DescribeInstanceInformation` drops any instance whose SSM agent isn't
   `Online` (`SendCommand` would hang or fail on those anyway).
2. A single `SendCommand` targets all the remaining instances.
3. `ListCommandInvocations` is polled on a fixed interval until every
   invocation finishes or the deadline passes, and we act on whatever
   results we have by then.

The per-instance outcome is unchanged: if the check reports `NOT_RUNNING` or the invocation `Failed`, `SetInstanceHealth` marks the instance
`Unhealthy`; `MARKER_EXISTS` is skipped.


template.yaml
```diff
@@ -267,7 +267,9 @@ Resources:
                     - !Sub "arn:aws:ssm:${AWS::Region}::document/AWS-RunShellScript"
                     - !Sub "arn:aws:ssm:${AWS::Region}::document/AWS-RunPowerShellScript"
                 - Effect: Allow
-                  Action: ssm:SendCommand
+                  Action:
+                    - ssm:SendCommand
+                    - ssm:ListCommandInvocations
                   Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
```
JoeColeman95: `ssm:ListCommandInvocations` would need to be handled the same way that `ssm:GetCommandInvocation` is below, it requires `Resource: "*"` and shouldn't be batched with `ssm:SendCommand` due to the crossover.

TLDR: These Actions won't work together unless the Resource is set to *

ref: https://docs.aws.amazon.com/service-authorization/latest/reference/list_awssystemsmanager.html

scaler/asg.go
```diff
@@ -122,127 +122,211 @@ esac
...
+			if inv.InstanceId == nil {
+				continue
+			}
+			results[*inv.InstanceId] = inv
+			switch inv.Status {
+			case ssmTypes.CommandInvocationStatusPending,
+				ssmTypes.CommandInvocationStatusInProgress,
+				ssmTypes.CommandInvocationStatusDelayed:
+				allTerminal = false
+			}
```
JoeColeman95: Cancelling is non-terminal but isn't in this set. There is an edge case if a cancelling invocation appears, the poll loop sees no non-terminal entries and exits early, the result handling loop would then fall through `isDangling` and logs `agent service is running` which is misleading.

```suggestion
			switch inv.Status {
			case ssmTypes.CommandInvocationStatusPending,
				ssmTypes.CommandInvocationStatusInProgress,
				ssmTypes.CommandInvocationStatusDelayed,
				ssmTypes.CommandInvocationStatusCancelling:
				allTerminal = false
			}
```

scaler/asg.go
```diff
@@ -122,127 +122,211 @@ esac
...
+			continue
 		}
-
-		checkOutput, err := ssmSvc.SendCommand(ctx, checkInput)
-		if err != nil {
-			log.Printf("[Elastic CI Mode] Warning: Could not send check command to instance %s: %v", instanceID, err)
+		switch inv.Status {
+		case ssmTypes.CommandInvocationStatusPending,
+			ssmTypes.CommandInvocationStatusInProgress,
+			ssmTypes.CommandInvocationStatusDelayed:
```
JoeColeman95: Same here 
```suggestion
		case ssmTypes.CommandInvocationStatusPending,
			ssmTypes.CommandInvocationStatusInProgress,
			ssmTypes.CommandInvocationStatusDelayed,
			ssmTypes.CommandInvocationStatusCancelling:
```

JoeColeman95 (changes requested):
The batching introduced here interacts with `MaxDanglingInstancesToCheck` param that's currently unbound in a way that the old per instance loop didn't. The default of 5 is safe, but anyone who chooses to set it to >50 will hit silent failures as `ListCommandInvocations` returns 50 per page without `NextToken` handling. So in this case, the early exit `len(results) >= expected` would never fire, and every check would spin for 60s.

We should either set `MaxValue: 50` 🩹,  or chunk `SendCommand` into <= 50 instance batches and paginate `ListCommandInvocations`.


Also, there's no test for `checkAndMarkUnhealthy` itself, there's tests added for the two helpers but the function that wires them together here is untested. This should be a quick win.

template.yaml
scadu: Ah, good point. I did make a manual policy change while testing which indeed was not bound to any Resources.
Addressed in [`7bd2171` (this PR)](https://github.com/buildkite/buildkite-agent-scaler/pull/314/commits/7bd21716b7739d12afd8ca083f54c3e8722a959c)

scaler/asg.go
scadu: Addressed in [`ffdd69a` (this PR)](https://github.com/buildkite/buildkite-agent-scaler/pull/314/commits/ffdd69a22ef2191df65672b2da7da2b23ec1d283)

scaler/asg.go
scadu: Addressed in [`ffdd69a` (this PR)](https://github.com/buildkite/buildkite-agent-scaler/pull/314/commits/ffdd69a22ef2191df65672b2da7da2b23ec1d283)

scadu:
> The batching introduced here interacts with `MaxDanglingInstancesToCheck` param that's currently unbound in a way that the old per instance loop didn't. The default of 5 is safe, but anyone who chooses to set it to >50 will hit silent failures as `ListCommandInvocations` returns 50 per page without `NextToken` handling. So in this case, the early exit `len(results) >= expected` would never fire, and every check would spin for 60s.
> 
> We should either set `MaxValue: 50` 🩹, or chunk `SendCommand` into <= 50 instance batches and paginate `ListCommandInvocations`.
> 
> Also, there's no test for `checkAndMarkUnhealthy` itself, there's tests added for the two helpers but the function that wires them together here is untested. This should be a quick win.

Oof, thanks for flagging this @JoeColeman95. Looks like I do have a draft with pagination implemented, but somehow ended up not committing it into the tree. Will test it against my stack and get it out.

JoeColeman95 (approved):

PR #332, opened 2026-07-31
Support full ARNs for SSM parameter and KMS key (cross-account agent token)

lizrabuya:
## What

Allow `BUILDKITE_AGENT_TOKEN_SSM_KEY` and `BuildkiteAgentTokenParameterStoreKMSKey` to be full ARNs, so the scaler can read the agent token from a parameter stored in a different AWS account.

Template changes (`template.yaml`):

- Two new `Conditions` — `IsAgentTokenARN` and `IsKMSKeyARN` — detect an ARN by splitting on `:` and checking whether the first segment is `arn`. This also covers non-`aws` partitions (`arn:aws-us-gov:...`, `arn:aws-cn:...`), which the previous hard-coded `arn:aws:` prefix wouldn't have.
- The `ssm:GetParameter` and `kms:Decrypt` IAM `Resource`s now use `!If` to pass the raw ref through when it's an ARN, or wrap it with `!Sub` as before when it's a path / key ID.
- Parameter `Description`s updated to document that ARNs are accepted.

No Lambda code change needed — `scaler.RetrieveFromParameterStore` (`scaler/ssm.go:10`) already passes the raw key straight into `GetParameter`'s `Name`, which the SDK accepts as either a path or an ARN.

README additions:

- Document that `BUILDKITE_AGENT_TOKEN_SSM_KEY` accepts a path or a full parameter ARN.
- Document that `BuildkiteAgentTokenParameterStoreKMSKey` accepts a key ID or a full key ARN, and call out that bare `alias/...` values are **not** supported (the template wraps non-ARN input as `:key/${...}`, which is invalid for aliases) — workaround is to pass the full alias ARN.
- Note that cross-account access also requires the owning account to attach a resource policy to the SSM parameter and a key policy to the KMS key granting `ssm:GetParameter` / `kms:Decrypt` to this Lambda's execution role. IAM permissions on the Lambda side are necessary but not sufficient.

## Why

Customers running the scaler in one AWS account want to point at a shared agent token parameter (and its encrypting KMS key) in a different account. Previously the template hard-coded `arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter${...}` and `arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${...}`, which pinned both resources to the same account/region as the Lambda.

## Notes on the design

- The ARN check (`!Equals [ !Select [0, !Split [":", !Ref X]], "arn" ]`) is coarse but adequate for the values we expect (path / key ID / alias / ARN).
- `IsKMSKeyARN` is evaluated at deploy time even when `UseKmsKeyForParameterStore` is false. With the default empty value, `!Split [":", ""]` yields `[""]` and the condition resolves cleanly to `false`, so the empty-default case is safe.
- No alias branch (`IsKMSAlias` on `alias/`) — bare aliases are a rare path and the README workaround (pass full alias ARN) is a reasonable trade-off.

## Testing

- Manual e2e: deployed the template with same-account path + key ID (existing behaviour), same-account full ARNs, and cross-account full ARNs against a `SecureString` parameter in a second AWS account. Verified the Lambda successfully reads and decrypts the token in each case.
- No automated CloudFormation tests exist in this repo, so this is a template-only change relying on manual verification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,12 @@ places this value directly in the `kms:Decrypt` IAM policy's `Resource` field, a
 key ARN there; an alias ARN in `Resource` grants no permissions. Resolve the alias to its
 underlying key ARN (via `aws kms describe-key --key-id alias/...`) and pass that instead.
 
-Cross-account access also requires configuration in the account that *owns* the parameter and key: attach a resource policy to the SSM parameter granting `ssm:GetParameter` to this Lambda's execution role, and a key policy on the KMS key granting `kms:Decrypt` to the same principal. The IAM permissions on the Lambda side are necessary but not sufficient.
+Cross-account access also requires configuration in the account that *owns* the parameter and key. The IAM permissions on the Lambda side are necessary but not sufficient:
+
+- The parameter must be **Advanced tier** — Standard-tier parameters (the default) cannot be shared cross-account. Create with `--tier Advanced` or upgrade an existing parameter (note: Advanced tier incurs a per-parameter monthly cost).
+- If the parameter is a `SecureString`, it must be encrypted with a **customer-managed KMS key** — the AWS-managed `alias/aws/ssm` key cannot be shared across accounts.
+- Attach a resource policy to the SSM parameter granting `ssm:GetParameter` to this Lambda's execution role.
+- Attach a key policy on the customer-managed KMS key granting `kms:Decrypt` to the same principal.
 
 ```bash
 aws lambda create-function \
--- a/README.md
+++ b/README.md
@@ -105,6 +105,8 @@ same-account use, or a full SSM parameter ARN (e.g., `arn:aws:ssm:us-east-1:1234
 to read a parameter in a different AWS account. For encrypted (`SecureString`) parameters, pass the
 full KMS key ARN via `BuildkiteAgentTokenParameterStoreKMSKey` so the Lambda can decrypt cross-account.
 
+Cross-account access also requires configuration in the account that *owns* the parameter and key: attach a resource policy to the SSM parameter granting `ssm:GetParameter` to this Lambda's execution role, and a key policy on the KMS key granting `kms:Decrypt` to the same principal. The IAM permissions on the Lambda side are necessary but not sufficient.
+
 ```bash
 aws lambda create-function \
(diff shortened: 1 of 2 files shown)


tomowatt (approved):
LGTM

README.md
```diff
@@ -100,6 +100,21 @@ If `BUILDKITE_AGENT_TOKEN_SSM_KEY` is set, the token will be read from
...
+
+`BuildkiteAgentTokenParameterStoreKMSKey` accepts a key ID (e.g., `abcd1234-...`) or a full key ARN
+(e.g., `arn:aws:kms:us-east-1:123456789012:key/abcd1234-...`). KMS aliases are **not** supported in
+either form — neither a bare alias (e.g., `alias/buildkite-token`) nor a full alias ARN
+(e.g., `arn:aws:kms:us-east-1:123456789012:alias/buildkite-token`) will work. The template
+places this value directly in the `kms:Decrypt` IAM policy's `Resource` field, and IAM requires a
+key ARN there; an alias ARN in `Resource` grants no permissions. Resolve the alias to its
+underlying key ARN (via `aws kms describe-key --key-id alias/...`) and pass that instead.
+
+Cross-account access also requires configuration in the account that *owns* the parameter and key: attach a resource policy to the SSM parameter granting `ssm:GetParameter` to this Lambda's execution role, and a key policy on the KMS key granting `kms:Decrypt` to the same principal. The IAM permissions on the Lambda side are necessary but not sufficient.
```
scadu: Could we mention that cross-account Parameter Store sharing only works with Advanced-tier parameters? `SecureString` parameters also need a customer-managed KMS key, since AWS-managed keys can't be shared. Standard parameters are the default, so without this note someone following these instructions can still get stuck after adding the policies. AWS lists both requirements in the [shared-parameter prerequisites](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-shared-parameters.html).

scadu (commented):
Nice fix — passing full ARNs through while preserving the existing path and key ID behavior looks right. I left one comment about documenting the AWS prerequisites for cross account Parameter Store sharing. Once that's covered, this looks good to me.

scadu (approved):

PR #337, opened 2026-08-20
Improve error message for 401 Unauthorized from Buildkite metrics API

### LINKED ISSUE #58: Improve error handling for HTTP 401 Unauthorized responses from Buildkite
These are [logged](https://github.com/buildkite/buildkite-agent-scaler/blob/8f7d8e2caca3808a75a33164a0720a5dae479286/buildkite/buildkite.go#L122-L124) but don’t provide much detail, ideally the error message would include a title, description, and a recovery suggestion.

The title should cover what the Lambda was attempting to do that failed e.g. "Couldn’t retrieve Buildkite metrics for the `queue` queue"

The description should provide more detail on what went wrong when trying to "do" the title e.g. "The Buildkite Agent token retrieved from (ENV VAR / SSM Parameter) was rejected"

The recovery suggestion should describe what action the user can take to fix the error e.g. "Retrieve or generate a new Buildkite Agent token from https://buildkite.com/organizations/-/agents and update the (ENV VAR / SSM Parameter) with the value".

CarlosZambonii:
## What

Fixes #58. When the Buildkite Agent Metrics API rejects the agent token (HTTP 401), the error message now includes:

- **Title**: what the scaler was attempting to do (e.g. `couldn't retrieve Buildkite metrics for the `queue` queue`)
- **Description**: that the token was rejected, naming exactly where it was retrieved from (env var, SSM parameter, or CLI flag) based on the actual runtime configuration
- **Recovery suggestion**: a pointer to https://buildkite.com/organizations/-/agents and instructions to update the same source

Previously it just logged a generic `GET <url>: 401 Unauthorized`, which several users found confusing (see comments on #58).

## How

- Added an `AgentTokenSource` field to `buildkite.Client`, set by both the CLI (`main.go`) and the Lambda handler (`lambda/main.go`) to describe where the token came from at runtime.
- Added a dedicated `http.StatusUnauthorized` branch in `queryMetrics` that builds the three-part message using that field. Non-401 non-200 responses keep their original generic error (unchanged).
- Added test coverage for both the SSM-parameter and environment-variable token sources.

## Testing

```
go build ./...   # pass
go vet ./...      # pass
gofmt -l ./        # clean
go test ./...     # all packages pass
```

### DIFF
--- a/buildkite/buildkite.go
+++ b/buildkite/buildkite.go
@@ -15,19 +15,29 @@ import (
 
 const (
 	PollDurationHeader = "Buildkite-Agent-Metrics-Poll-Duration"
+
+	// defaultAgentTokenSource is used in error messages when the caller
+	// hasn't set AgentTokenSource to describe where the token came from.
+	defaultAgentTokenSource = "configured source"
 )
 
 type Client struct {
 	Endpoint   string
 	AgentToken string
-	UserAgent  string
+	// AgentTokenSource describes where AgentToken was retrieved from, e.g.
+	// "BUILDKITE_AGENT_TOKEN environment variable" or `SSM parameter
+	// "/my/param"`. It's used to give a more actionable error message if
+	// the token is rejected by the Buildkite Agent Metrics API.
+	AgentTokenSource string
+	UserAgent        string
 }
 
 func NewClient(agentToken, agentEndpoint string) *Client {
 	return &Client{
-		Endpoint:   agentEndpoint,
-		UserAgent:  fmt.Sprintf("buildkite-agent-scaler/%s", version.VersionString()),
-		AgentToken: agentToken,
+		Endpoint:         agentEndpoint,
+		UserAgent:        fmt.Sprintf("buildkite-agent-scaler/%s", version.VersionString()),
+		AgentToken:       agentToken,
+		AgentTokenSource: defaultAgentTokenSource,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


JoeColeman95 (changes requested):
Hey @CarlosZambonii,

Appreciate the contribution! 🎉 

The changes overall look good, but there's a few things that I'd suggest keeping, firstly we are removing the URL for the endpoint that's being hit with these changes, I'd suggest that we return this so that it's clear where the error is coming from, as we just blanket every request as a bad token when it may be an issue with DNS in-between, for example.

Secondly, we're throwing away the response body here but serving this is likely better than serving a static response, as this will state if the token is missing or was rejected.

Finally, `BUILDKITE_AGENT_TOKEN_SSM_KEY` is only local to the agent scaler in this case, which is passed down via CloudFormation/TF from the Elastic CI Stack. It may be worth removing this entry to avoid confusion. The parameter path should suffice in my opinion 🙂

CarlosZambonii:
Thanks so much for the thorough review, really appreciate it! 🙌

Pushed a fix covering all three points:

- Error message includes the request URL again
- API response body is now folded into the message (helps tell "token missing" apart from "token rejected")
- Dropped the internal `BUILDKITE_AGENT_TOKEN_SSM_KEY` name from the SSM-sourced message — just the parameter path now

`go build`, `go vet`, `go test ./...`, `gofmt -l` all clean. Let me know if anything else needs adjusting!

Side note, no action needed unless you think it's worth it: while going through it I noticed the new error-body handling doesn't cap the response size and the tests only cover the JSON-with-message case — happy to add a size limit + a couple more test cases in a follow-up if that seems useful.

buildkite/buildkite.go
```diff
@@ -131,3 +152,26 @@ func (c *Client) queryMetrics(ctx context.Context, into interface{}, queue strin
 
 	return pollDuration, json.NewDecoder(res.Body).Decode(into)
 }
+
+// responseErrorDetail reads res.Body and folds it into a human-readable
+// status string, e.g. `401 Unauthorized: Eeep! You forgot to pass an agent
+// registration token`. Callers should not read res.Body afterwards.
+func responseErrorDetail(res *http.Response) string {
+	body, err := io.ReadAll(res.Body)
```
lizrabuya: This handler reads an untrusted response body without a size limit, potentially causing excessive memory use and oversized CloudWatch logs. You can cap or truncate the body, or make the rejection reason concise e.g. make it assert that the token 'was rejected' and advise  generating a replacement, even when the response comes from an intermediary rather than Buildkite.

lizrabuya (changes requested):

PR #345, opened 2026-08-26
Handle draining agents during dangling cleanup

scadu:
## Why

The termination marker only tells us SIGTERM was sent; it doesn't tell us whether the agent is still running. Dangling cleanup could therefore keep skipping a marked instance after its agent had stopped.

## What

- Treat a marked, running agent as draining.
- Mark instances with stopped agents as unhealthy for replacement.
- Skip failed or ambiguous SSM checks rather than guessing.
- Avoid repeated scale-in attempts while the ASG is already converging, and clarify the related logs.

Tested end-to-end on an Elastic CI Stack: a busy agent received one SIGTERM and drained normally, while dangling cleanup replaced a stopped agent without sending another SIGTERM.

### DIFF
--- a/scaler/asg.go
+++ b/scaler/asg.go
@@ -116,16 +116,29 @@ if ($AgentStatus -match "SERVICE_RUNNING") {
 	// Default to Linux
 	return `#!/bin/bash
 # Linux check command
-if [ -f /tmp/buildkite-agent-termination-marker ]; then
-  echo "MARKER_EXISTS: Instance is already marked for termination"
+if ! SERVICE_STATE=$(systemctl show buildkite-agent --property=ActiveState --property=MainPID 2>&1); then
+  echo "UNKNOWN: $SERVICE_STATE"
   exit 0
 fi
 
-ACTIVE_STATE=$(systemctl show buildkite-agent -p ActiveState | cut -d= -f2)
-case "$ACTIVE_STATE" in
-  "active"|"activating") echo "RUNNING" ;;
-  *) echo "NOT_RUNNING: $ACTIVE_STATE" ;;
+ACTIVE_STATE=$(printf '%s\n' "$SERVICE_STATE" | sed -n 's/^ActiveState=//p')
+MAIN_PID=$(printf '%s\n' "$SERVICE_STATE" | sed -n 's/^MainPID=//p')
+case "$ACTIVE_STATE:$MAIN_PID" in
+  :*|*:|*:*[!0-9]*)
+    echo "UNKNOWN: ActiveState=$ACTIVE_STATE MainPID=$MAIN_PID"
+    exit 0
+    ;;
 esac
+
+if [ "$MAIN_PID" != "0" ] || [ "$ACTIVE_STATE" = "activating" ]; then
+  if [ -f /tmp/buildkite-agent-termination-marker ]; then
+    echo "DRAINING: ActiveState=$ACTIVE_STATE MainPID=$MAIN_PID"
+  else
+    echo "RUNNING"
+  fi
+else
+  echo "NOT_RUNNING: ActiveState=$ACTIVE_STATE"
+fi
 `
 }
 
@@ -255,20 +268,20 @@ func (a *ASGDriver) checkAndMarkUnhealthy(
 	ssmSvc ssmCheckAPI,
 	asgSvc asgHealthAPI,
 	platform string,
-) (markedUnhealthyCount int, checkedCount int, firstError error) {
+) (markedUnhealthyCount int, checkedCount int, drainingCount int, firstError error) {
 	if len(instances) == 0 {
-		return 0, 0, nil
+		return 0, 0, 0, nil
 	}
 
 	onlineIDs, err := filterOnlineSSMInstances(ctx, ssmSvc, instances)
 	if err != nil {
-		return 0, 0, fmt.Errorf("DescribeInstanceInformation failed: %w", err)
+		return 0, 0, 0, fmt.Errorf("DescribeInstanceInformation failed: %w", err)
 	}
 	if offline := len(instances) - len(onlineIDs); offline > 0 {
 		log.Printf("[Elastic CI Mode] SSM agent not online for %d of %d instance(s); skipping those", offline, len(instances))
 	}
 	if len(onlineIDs) == 0 {
-		return 0, 0, nil
+		return 0, 0, 0, nil
 	}
 
 	documentName := "AWS-RunShellScript"
@@ -288,7 +301,7 @@ func (a *ASGDriver) checkAndMarkUnhealthy(
 			Comment:      aws.String("Check if buildkite-agent service is running"),
 		})
 		if err != nil {
-			return 0, 0, fmt.Errorf("SendCommand failed: %w", err)
+			return 0, 0, 0, fmt.Errorf("SendCommand failed: %w", err)
 		}
 		commandIDs = append(commandIDs, *sendOut.Command.CommandId)
 	}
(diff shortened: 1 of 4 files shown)


scaler/scaler.go
```diff
@@ -240,6 +240,10 @@ func (s *Scaler) Run(ctx context.Context) (time.Duration, error) {
 	if instanceCount == 0 {
 		instanceCount = asg.DesiredCount
 	}
+	if asg.DesiredCount == desired && instanceCount != desired {
+		log.Printf("ASG is converging to desired capacity %d (currently %d running instances)", desired, instanceCount)
+		return metrics.PollDuration, nil
+	}
```
petetomasik, line 246: This early return isn't tested in `scaler_test.go`. `TestScalingInWithoutError` and `TestScalingOutWithoutError` never set `actualCapacity`, so `instanceCount` always equals `asg.DesiredCount` and this condition can never be true.

I traced all nine combinations of (desired vs. instanceCount, desired vs. asg.DesiredCount) and the refactor is correct: exactly one case changes behaviour, `instanceCount > desired && asg.DesiredCount == desired`, which is the no-op scale-in you're removing. My concern is that the correctness of line 256 (`Action: Scale in from ...`, now unconditional) depends on this block having already consumed the `asg.DesiredCount == desired` case. That invariant is non-local and unasserted, so if anyone later moves this block below `if desired > instanceCount`, the `desired > instanceCount && asg.DesiredCount == desired` case falls through to `s.scaleIn` and issues a capacity reduction on a path that should be a no-op, with nothing failing in CI.

`asgTestDriver` already has an `actualCapacity` field, so this is two table entries rather than new scaffolding:

- `asgDesired: 3, actualCapacity: 5`, desired computing to 3: expect no `SetDesiredCapacity` and no SIGTERMs (the converging case this commit adds).
- `asgDesired: 5, actualCapacity: 3`, desired computing to 3: expect `scaleIn` still fires, proving the early return didn't swallow a real scale-in.

The second is the one that would catch a reordering regression. Worth noting the asymmetry too: `asg.go` got 117 lines of new tests here and the change to the core scaling decision got none.

scaler/asg.go
```diff
@@ -623,28 +633,13 @@ func (a *ASGDriver) CleanupDanglingInstances(ctx context.Context, minimumInstanc
...
-	if totalMarkedUnhealthy > 0 {
-		log.Printf("[Elastic CI Mode] Dangling instance check: marked %d of %d checked instance(s) as unhealthy (%d skipped due to errors)", totalMarkedUnhealthy, totalChecked, skipped)
-	} else if skipped > 0 {
-		log.Printf("[Elastic CI Mode] Dangling instance check complete: %d of %d instance(s) checked and healthy, %d skipped due to errors", totalChecked, len(instancesForSSMCheck), skipped)
-	} else {
-		log.Printf("[Elastic CI Mode] Dangling instance check complete: all %d instance(s) healthy", totalChecked)
-	}
+	log.Printf("[Elastic CI Mode] Checking %d %s instance(s) for dangling agents: %v", len(instancesForSSMCheck), platform, instancesForSSMCheck)
+	markedUnhealthy, checked, draining, checkErr := a.checkAndMarkUnhealthy(ctx, instancesForSSMCheck, ssmClient, asgClient, platform)
+	healthy := checked - draining - markedUnhealthy
```
petetomasik: This is correct today only because `checkedCount++` appears in exactly three places (lines 349, 353 and 379) and one of them is paired with `markedUnhealthyCount++`. Any future `checkedCount++` added anywhere in the loop silently corrupts the `healthy` figure in this operator-facing log line, and there's no test that would catch it because nothing asserts on the log output.

Cheaper to return what the caller actually wants than to reconstruct it: either a fourth named return for `healthyCount`, or fold the four counters into a small result struct so the accounting stays in one place next to the classification that produces it.

Related, and the reason I noticed: `checkedCount` no longer means "checked". Line 379 puts the increment inside the `SetInstanceHealth` success branch, so an instance with a conclusive `NOT_RUNNING` whose health-mark call fails is excluded from `checked` and shows up in `skipped` instead. That's the right bucket for it, but the name now says something different from what it counts.

scaler/asg_test.go
```diff
@@ -313,6 +316,62 @@ func TestGetCheckCommand(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			marker := filepath.Join(t.TempDir(), "termination-marker")
+			if tc.marker {
+				if err := os.WriteFile(marker, nil, 0o600); err != nil {
+					t.Fatalf("write termination marker: %v", err)
+				}
+			}
+
+			script := strings.ReplaceAll((&ASGDriver{}).getCheckCommand("linux"), "/tmp/buildkite-agent-termination-marker", marker)
```
petetomasik: This reaches into the generated script to swap the marker path, which couples the test to a string literal it doesn't own. If the path in `getCheckCommand` ever changes, `ReplaceAll` matches nothing, the script falls back to probing the real `/tmp/buildkite-agent-termination-marker` on whatever machine is running the suite, and the `draining` and `marked but stopped` subtests start reporting on host state rather than the fixture. It fails loudly rather than silently, so this isn't urgent, but the failure gives no hint about the cause.

scaler/asg_test.go
```diff
@@ -313,6 +316,62 @@ func TestGetCheckCommand(t *testing.T) {
...
+func TestLinuxCheckCommand(t *testing.T) {
+	binDir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(binDir, "systemctl"), []byte(`#!/bin/sh
+if [ -n "$SYSTEMCTL_ERROR" ]; then
+  echo "$SYSTEMCTL_ERROR"
+  exit 1
+fi
+printf 'ActiveState=%s\nMainPID=%s\n' "$ACTIVE_STATE" "$MAIN_PID"
+`), 0o755); err != nil {
+		t.Fatalf("write systemctl stub: %v", err)
```
petetomasik, line 328: Good table, and testing the shell by executing it against a stubbed `systemctl` is the right approach. The six cases here cover `active`, `activating`, `inactive`, systemctl failure and a malformed PID, but they don't cover the three states whose classification this PR actually changes:

- `deactivating` with a live `MainPID` and no marker: was `NOT_RUNNING`, now `RUNNING`.
- `failed`: reaches `NOT_RUNNING` via `MainPID=0` rather than via the old `case *)` fallthrough.
- `active` with `MainPID=0`: was `RUNNING`, now `NOT_RUNNING`.

The states that kept their classification are the ones least likely to regress. Adding those three is three more lines in the existing table and pins the behaviour change this commit is making, including the `Type=simple` assumption the `active`/`MainPID=0` case now depends on.

Also worth a case for `deactivating` *with* a marker, since that's the normal graceful scale-in window and the one path an operator will actually watch: it should report `DRAINING`, and right now nothing asserts that.

petetomasik (commented):
Some comments, mainly about test coverage.

scaler/asg_test.go
scadu: Addressed in 7e0e311

scaler/asg.go
scadu: Addressed in 0a3b574

scaler/asg_test.go
scadu: Addressed in 83e5a16

scaler/scaler.go
scadu: Addressed in b08a0ef

petetomasik (approved):

PR #350, opened 2026-08-31
fix: stop Elastic CI scale-in from double-decrementing desired capacity

kurigashvili-benchling:
## Summary

We hit this on scaler 1.12.0. Elastic CI scale-in sent SSM `systemctl stop buildkite-agent`, which runs `ExecStopPost=/usr/local/bin/terminate-instance` (`TerminateInstanceInAutoScalingGroup` with `ShouldDecrementDesiredCapacity=true`). The scaler also called `SetDesiredCapacity` right after SIGTERM. That is two decrements.

**`SendSIGTERMToAgents` did not serialize those two decrements.** In source order it ran *before* `SetDesiredCapacity`, but it only `ssm:SendCommand`. That returns when AWS accepts the command, not when `systemctl stop` finishes, not when `ExecStopPost` / `terminate-instance` runs, and not when desired capacity is decremented. The instance-side decrement is async and can (and did) land *after* `SetDesiredCapacity`. Putting SIGTERM first in the function did not guarantee terminate-and-decrement completed first.

Once desired hit `MinSize`, PostStop failed, agents died, and the instances stayed `InService`:

```
terminate-instance[10437]: aws: [ERROR]: An error occurred (ValidationError) when calling the
TerminateInstanceInAutoScalingGroup operation: Currently, desiredSize equals minSize (1).
Terminating instance without replacement will violate group's min size constraint. Either set
shouldDecrementDesiredCapacity flag to false or lower group's min size.
```

Screenshot 2026-08-31 at 11 52 56 AM

Screenshot 2026-08-31 at 11 53 11 AM

A newer stack `terminate-instance` would have unblocked **that min-size error**. [v6.71.0](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v6.71.0/packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance) falls back to terminate **without** decrement when `--should-decrement-desired-capacity` fails (desired already at min). The instances would then terminate instead of sitting `InService` with a dead agent.

It does **not** stop the double decrement. Without this PR, the scaler still `SetDesiredCapacity` and each PostStop still decrements. Scaling to a target **above** min still overshoots: `SetDesiredCapacity(10)` lands first, then each PostStop walks `10 → 9 → …`. This PR is required to make the scaler the sole owner of desired capacity.

#345 already teaches the dangling probe to mark stopped agents unhealthy. That does not stop the double decrement.

## Approach

The scaler must not `systemctl stop` the agent. That always runs `ExecStopPost`. There is no “stop the agent without `terminate-instance`” on these AMIs.

Elastic CI instances launch with `NewInstancesProtectedFromScaleIn`, so lowering desired capacity alone does not remove them. `SetInstanceHealth(Unhealthy)` bypasses protection.

This PR:

1. **`SetDesiredCapacity(target)` first.** Marking unhealthy while desired is still high makes the ASG launch replacements.
2. **`SetInstanceHealth(Unhealthy)` on the N oldest InService extras** (lambda, same API dangling cleanup already uses). The ASG terminates them down to the new desired count; no replacements. Pending and Terminating instances are excluded so a retry after a partial health-mark does not re-target draining instances and skip remaining extras.
3. **Leave drain to the stack lifecycle hook** (`stop-agent-gracefully`). Do not SSM-stop the agent.

If desired is already at target but extras are still `InService`, skip cooldown and only mark them unhealthy (retry after a partial health-mark failure).

## Why the lifecycle hook does not double-decrement

`stop-agent-gracefully` itself does not decrement desired capacity. It only stops the agent and waits for it to exit.

For a normal (non-spot) ASG termination it does `systemctl stop buildkite-agent`. That still hits systemd `ExecStopPost=/usr/local/bin/terminate-instance`, which tries `TerminateInstanceInAutoScalingGroup --should-decrement-desired-capacity`.

In this PR's path that call should not actually drop capacity. By then the scaler has already set desired capacity, the instance is `Unhealthy`, and the ASG has put it in `Terminating:Wait`. `TerminateInstanceInAutoScalingGroup` is meant for `InService` instances, so that API typically errors. The script then tries terminate-without-decrement, then `SetInstanceHealth(Unhealthy)` — both no-ops on an instance the ASG is already terminating.

That is the difference from the old scaler SIGTERM path: the instance was still `InService`, so `ExecStopPost` **did** decrement, and the scaler also set desired capacity (double decrement).

Capacity decrement is owned by the scaler's `SetDesiredCapacity`. The instance may still attempt it via `ExecStopPost`; it should not succeed.

## Test plan

- [x] `go test ./...` / `go vet ./...`
- [x] Elastic CI scale-in calls `SetDesiredCapacity` then `SetInstanceHealth` on the oldest extras
- [x] `SetInstanceHealth` is skipped if `SetDesiredCapacity` fails (would launch replacements)
- [x] When desired is already at target but actual is high, extras are marked unhealthy without another desired-capacity change
- [x] `Describe` instance IDs are InService only (Pending/Terminating excluded)
- [ ] Manual: scale a burst down to `MinSize`, confirm extras terminate and desired does not walk below min
- [ ] Manual: scale in to a target **above** `MinSize`, confirm desired stays at that target

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,8 @@ additionally requires:
 
 * `autoscaling:SetInstanceHealth`
 * `autoscaling:SetInstanceProtection` — Elastic CI Stack instances are launched protected from
-  scale-in; dangling (agentless) instances and scale-in candidates that cannot self-terminate after
-  SIGTERM must have protection cleared before the ASG will terminate or replace them
+  scale-in; dangling (agentless) instances and all-fallback scale-in batches that cannot
+  self-terminate must have protection cleared before the ASG will terminate or replace them
 * `ssm:SendCommand`, `ssm:ListCommandInvocations`, `ssm:DescribeInstanceInformation`
 * `ec2:DescribeInstances`, `ec2:DescribeInstanceStatus`, `ec2:TerminateInstances`
 
--- a/README.md
+++ b/README.md
@@ -93,9 +93,8 @@ With `ELASTIC_CI_MODE` enabled, the lambda reclaims instances whose agent has st
 additionally requires:
 
 * `autoscaling:SetInstanceHealth`
-* `autoscaling:SetInstanceProtection` — Elastic CI Stack instances are launched protected from
-  scale-in; dangling (agentless) instances and all-fallback scale-in batches that cannot
-  self-terminate must have protection cleared before the ASG will terminate or replace them
+* `autoscaling:SetInstanceProtection` — required before the ASG will terminate or replace a
+  protected Elastic CI instance (dangling reclaim, and scale-in batches that cannot self-terminate)
 * `ssm:SendCommand`, `ssm:ListCommandInvocations`, `ssm:DescribeInstanceInformation`
 * `ec2:DescribeInstances`, `ec2:DescribeInstanceStatus`, `ec2:TerminateInstances`
 
--- a/README.md
+++ b/README.md
@@ -89,6 +89,15 @@ the following IAM permissions:
 * `autoscaling:DescribeScalingActivities`
 * `autoscaling:SetDesiredCapacity`
 
+With `ELASTIC_CI_MODE` enabled, the lambda reclaims instances whose agent has stopped and gracefully
+scales in the ones it selects for termination, which additionally requires:
+
+* `autoscaling:SetInstanceHealth`
+* `autoscaling:SetInstanceProtection` — Elastic CI Stack instances are launched protected from
+  scale-in, and the ASG will not terminate or replace them until that protection is cleared
+* `ssm:SendCommand`, `ssm:ListCommandInvocations`, `ssm:DescribeInstanceInformation`
+* `ec2:DescribeInstances`, `ec2:DescribeInstanceStatus`, `ec2:TerminateInstances`
+
 Its handler is `bootstrap`, it uses a `provided.al2023` runtime and requires the following env vars:
 
 * `BUILDKITE_AGENT_TOKEN` or `BUILDKITE_AGENT_TOKEN_SSM_KEY`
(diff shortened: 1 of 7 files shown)


scadu:
Thanks for the detailed writeup @kurigashvili-benchling. Your analysis is right. The scaler calls `SetDesiredCapacity` while each instance's `ExecStopPost` also decrements, so every graceful scale-in removes capacity twice, and at `MinSize` that strands dead `InService` instances. We want to fix this too, but we'd like to go with #351 instead: it drops the scaler-side decrement so the instance stays the single owner of it, and keeps the existing drain path.

The concern with `SetInstanceHealth(Unhealthy)`: on the stock stack there's no termination lifecycle hook, so nothing drains the agent before the ASG terminates the instance. `stop-agent-gracefully` is wired up as the lifecycled handler, but the stack [configures lifecycled without an SNS topic](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/c18f1465/packer/linux/stack/conf/bin/bk-install-elastic-stack.sh#L602-L607), and lifecycled [only listens for ASG lifecycle events when one is set](https://github.com/buildkite/lifecycled/blob/0a944f09/daemon.go#L45-L48) — out of the box it only reacts to spot interruptions. So marking an instance unhealthy terminates it directly (no `Terminating:Wait`), and running jobs get killed.

Do you happen to have a custom termination lifecycle hook on your ASG? That would explain a clean drain in your validation.

#351 also batches the SSM graceful-stop into a single `SendCommand` and only returns `InService` instances as scale-in candidates, so terminating instances don't get re-selected.

kurigashvili-benchling:
@scadu You are correct. SetInstanceHealth(Unhealthy) is not good fit as it can terminate running jobs. I will close this PR. Lets go with https://github.com/buildkite/buildkite-agent-scaler/pull/351

Thank you.

PR #351, opened 2026-09-01
Batch graceful scale-in and drop the scaler-side decrement

scadu:
Fixes the double decrement on graceful scale-in, reported in #350.

Both sides were decrementing desired capacity: the scaler called `SetDesiredCapacity`, and the Elastic CI Stack instance also terminates itself with [`--should-decrement-desired-capacity`](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/c18f14658528df73af2189ffcee84774ec331faa/packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance#L9) after draining. Once the ASG hits `MinSize`, the instance's decrement fails and it just sits there `InService` with no agent — those dead instances pile up until someone cleans them up manually.

The instance already knows how to drain and remove itself, so the scaler now stays out of it:

- No more `SetDesiredCapacity` on the Linux graceful path — the instance owns the decrement. Windows is unchanged (no graceful stop there, so the scaler still decrements).
- Graceful-stop goes out as one batched SSM `SendCommand` (chunked at the API's 50-instance limit) instead of one call per instance with 30s of readiness polling each.
- `Describe` only returns `InService` instances now, so we stop re-selecting ones that are already terminating.
- Scale-in cooldown only starts when we actually dispatched something.

Tests cover the batch chunking, the Windows fallback, and the `InService` filtering.


scaler/scaler.go
```diff
@@ -556,6 +519,29 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
...
+				log.Printf("CRITICAL: [Elastic CI Mode] Failed to set desired capacity to %d for Windows instances: %v", desired, err)
+			} else {
+				s.scaleInParams.LastEvent = time.Now()
+			}
+			return
+		}
+
+		log.Printf("⚠️  Failed to send graceful-stop command to one or more instances: %v", err)
+	}
+	if dispatched > 0 {
```
petetomasik: When `filterOnlineSSMInstances` comes back empty, `sendSIGTERMToAgentsBatch` returns `(0, nil)` (asg.go:572-574). That lands here as `err == nil`, so we skip the warning branch, and `dispatched == 0`, so we skip `LastEvent`. Combined with no `SetDesiredCapacity` on the Linux path, this invocation does nothing at all, and `scaleIn` still returns `nil` at scaler.go:511, so `Run` and the Lambda both report success.

Trigger: any Elastic CI ASG whose instances can't reach SSM (instance profile missing `ssm:UpdateInstanceInformation`, private subnet without SSM VPC endpoints, dead SSM agent). There's no other exit either: `CleanupDanglingInstances` -> `checkAndMarkUnhealthy` runs the same filter and returns early at asg.go:285-287 for exactly those instances.

The second half is the operationally expensive one. Because `LastEvent` is never set, the cooldown guard at scaler.go:302 is skipped, so every `LAMBDA_INTERVAL` (10s default) re-runs the whole scale-in path, including `GetLastScalingInAndOutActivity` at scaler.go:334 with `MaxDescribeScalingActivitiesPages` force-set to -1 at scaler.go:327. That's a full activity-history walk under a 30s timeout, every ten seconds, forever.

Worth noting this is a regression even on the default protected-instance config, where neither the old nor the new code could actually remove the instance. The old code still called `SetDesiredCapacity`, which made `asg.DesiredCount == desired` and tripped the "ASG is converging" early return at scaler.go:243 on subsequent runs, so the scaler went quiet. Now it never does.

Suggestion: treat `dispatched == 0 && len(instanceIDs) > 0` as a failure. Return an error from `gracefullyScaleIn` and propagate it out of `scaleIn` so the invocation isn't reported as successful, and either fall back to `setDesiredCapacity` or at minimum set `LastEvent` so the cooldown gives us backoff.

scaler/scaler.go
```diff
@@ -460,41 +458,7 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
...
-		if sigTermSuccess > 0 {
-			log.Printf("✅ Successfully sent SIGTERM to %d instance(s)", sigTermSuccess)
-		}
-
-		log.Printf("[Elastic CI Mode] Updating ASG desired capacity to %d", desired)
-		if err := s.setDesiredCapacity(ctx, desired); err != nil {
-			log.Printf("CRITICAL: [Elastic CI Mode] Failed to set desired capacity to %d after sending SIGTERMs: %v. ASG might replace terminated instances.", desired, err)
-
-		}
+		s.gracefullyScaleIn(ctx, instancesForTermination, desired)
```
petetomasik: The `setDesiredCapacity(ctx, desired)` that used to sit right here was also load-bearing for the single-instance block immediately below (463-510). That block terminates via `directlyTerminateInstance` (scaler.go:619-629), which is a plain `ec2:TerminateInstances` with no capacity decrement. On master it worked because desired capacity had already been lowered one line earlier, so the ASG was at the new count when the terminate landed. Now desired is untouched, so the ASG just launches a replacement.

Trigger: Elastic CI mode, `DesiredCount == 1`, `MinSize == 0`, computed `desired == 0` (needed to satisfy `instancesToTerminate > 0` and enter the branch at :396), and the instance's SSM agent unreachable, which is exactly the condition the probe at :490 tests for. Sequence: `gracefullyScaleIn` dispatches 0 so no cooldown is recorded (see my comment on :540), the probe `SendCommand` errors, we take the "assuming dangling instance" path, `TerminateInstances` kills it, desired is still 1, ASG boots a replacement. Next runs return early on `asg.Pending > 0` at :258 until it's InService, then the cycle repeats. Unbounded terminate/replace churn where the old code reached 0 and stayed there.

Note scale-in protection doesn't save us here: it explicitly does not cover manual EC2 terminations, so the instance does die, we just pay to relaunch it.

Also worth checking: the branch condition itself widened. `len(current.InstanceIDs) == 1` used to mean "one instance in the ASG"; with the new InService filter in `autoscaleGroupDetails` it means "one InService instance", so this now also fires with 1 InService + 1 Terminating. In that shape a transient `SendCommand` throttle hard-kills an instance that's mid-drain on a real job.

Suggestion: call `setDesiredCapacity(ctx, desired)` before `directlyTerminateInstance`, or switch it to `TerminateInstanceInAutoScalingGroup` with `ShouldDecrementDesiredCapacity: true`. Guarding the block on `dispatched == 0` would also stop us probing an instance we just successfully signalled.

scaler/scaler.go
```diff
@@ -556,6 +519,29 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
 	}
 }
 
+func (s *Scaler) gracefullyScaleIn(ctx context.Context, instanceIDs []string, desired int64) {
+	// Elastic CI Stack agents self-terminate and decrement desired capacity
+	// after draining. Setting desired capacity here for Linux would decrement
```
petetomasik: This comment states the design decision, and I think it has a consequence outside this function that we should handle before merging: it silently kills the Elastic CI scale-in cooldown.

`GetLastScalingInAndOutActivity` (asg.go:493-499) only matches an activity whose `Cause` contains both `userRequestForChangingDesiredCapacity` ("a user request explicitly set group desired capacity changing the desired capacity", asg.go:31) and `shrinkingKey`. That first phrase is what `SetDesiredCapacity` emits. The instance-side terminate emits a different one. From the AWS docs:

    "At 2023-09-23T06:39:59Z instance i-026e4c9f62c3e448c was taken out of service in response to a user request, shrinking the capacity from 1 to 0."

It has `shrinkingKey` but not the required first phrase. Since the Linux path is now the only scale-in mechanism and never calls `SetDesiredCapacity`, no matching activity is ever recorded again. Two consequences:

1. scaler.go:334-352 always sees `lastScaleInActivity == nil`, so the ASG-activity cooldown never fires. Worse, each query now walks the entire history looking for a match that can't exist: the loop at asg.go:505-515 only breaks on `NextToken == nil`, and `MaxDescribeScalingActivitiesPages` is force-set to -1 in Elastic CI mode at scaler.go:327.

2. lambda/main.go:129-143 seeds `ScaleInParams.LastEvent` from the same query. Once the pre-upgrade `SetDesiredCapacity` activities age out of the retention window, `lastScaleIn` is the zero time, so `if !s.scaleInParams.LastEvent.IsZero()` at scaler.go:302 skips the cooldown check entirely on the first scale-in after every cold start. The 1h default we document in docs/elastic_ci_mode.md:104 silently stops being enforced.

Suggestion: teach the matcher the terminate-with-decrement cause shape as well, something like ("was taken out of service in response to a user request" + `shrinkingKey`), so both the in-run cooldown and the cold-start seeding keep working once the scaler is no longer the one changing desired capacity.

scaler/asg.go
```diff
@@ -567,21 +584,27 @@ fi
...
+	var sendErrors []error
+	batchNumber := 0
+	dispatched := 0
+	for instanceBatch := range slices.Chunk(onlineIDs, ssmMaxInstanceIDs) {
+		batchNumber++
+		_, err := ssmSvc.SendCommand(ctx, &ssm.SendCommandInput{
+			InstanceIds:  instanceBatch,
+			DocumentName: aws.String("AWS-RunShellScript"),
+			Parameters:   map[string][]string{"commands": {command}},
+			Comment:      aws.String("Gracefully stop Buildkite agent"),
```
petetomasik, line 623: This needs `MaxErrors: aws.String("100%")`. Batching 50 targets into one call changes the blast radius of a single bad instance, and the SSM defaults work against us here.

From the SendCommand API reference, `MaxErrors` defaults to `0`: "When the command fails one more time beyond the value of `MaxErrors`, the system stops sending the command to additional targets." And delivery isn't simultaneous. Per the Run Command guide, "the queueing system delivers the command to a single node and waits until the system acknowledges the initial invocation before sending the command to two more nodes", ramping exponentially up to `MaxConcurrency`.

Trigger: one instance passes the `PingStatus == Online` filter but is already gone. `PingStatus` reflects a cached heartbeat and lags real reachability by minutes, so a recently dead instance still reads Online. Its invocation fails early in the ramp, SSM stops delivering to the rest of the batch, and up to 49 instances never receive the graceful stop. Meanwhile `dispatched += len(instanceBatch)` on :603 has already counted all 50, so `gracefullyScaleIn` sets `LastEvent` and the 1h Elastic CI cooldown blocks a retry. Nothing polls `ListCommandInvocations` on this path (deliberately, and I agree with that), so the truncation is completely invisible.

This is new exposure from the batching rather than something we inherited: the old code sent one command per instance, so a failure was scoped to that instance. `checkAndMarkUnhealthy` at :298 has the same omission, but it polls invocations, so truncation surfaces there.

Related and cheaper to fix at the same time: `dispatched` is really "instances included in an accepted API call", not "instances signalled". The doc comment on `SendSIGTERMToAgentsBatch` says as much, but the caller treats it as the latter when it decides whether to start the cooldown. Worth either renaming it or gating the cooldown on `dispatched == len(instanceIDs)`.

petetomasik (changes requested):
Some findings we might need to address before merging.

scaler/scaler.go
scadu: Sorted out in e53674b

scaler/scaler.go
scadu: Addressed in 9fbdf9b and aa08a82

scaler/scaler.go
scadu: Should be covered in fe69840

scaler/scaler.go
```diff
@@ -393,116 +403,30 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
...
-
-		if current.DesiredCount <= 1 && len(current.InstanceIDs) == 1 {
+		gracefulScaleInErr := s.gracefullyScaleIn(ctx, instancesForTermination, desired)
+
+		// Only probe when nothing was reachable over SSM. A failed SSM or EC2
+		// API call tells us nothing about the instance, and the probe would
+		// most likely fail the same way before hard-killing an agent that
+		// might be mid-job.
+		failedGracefulHandoff := errors.Is(gracefulScaleInErr, errNoReachableScaleInCandidates)
+		singleInstanceASG := current.DesiredCount <= 1 && current.TotalCount == 1 && len(current.InstanceIDs) == 1
```
petetomasik: This gate leaves multi-instance ASGs with no way to scale in at all when SSM is unreachable, which is a convergence regression against master.

`singleInstanceASG` requires `current.TotalCount == 1`, so for any ASG with more than one instance the rescue below never runs. There's no other exit on the Linux path either: `sendSIGTERMToAgentsBatch` returns `(0, nil)` when `filterOnlineSSMInstances` comes back empty (asg.go:584-586), `gracefullyScaleIn` records `LastEvent` at scaler.go:514 and returns `errNoReachableScaleInCandidates`, and `SetDesiredCapacity` is never called.

Trigger: Elastic CI mode, `DesiredCount=5`, `MinSize=0`, all five InService, computed `desired=2`, and none of the candidates reachable over SSM. Same conditions as before: instance profile missing `ssm:UpdateInstanceInformation`, private subnet without SSM VPC endpoints, or a dead SSM agent. `instancesToTerminate` is 3, `oldestInstances` picks 3, the online filter drops all of them, and we return the sentinel. Next poll hits the cooldown check at scaler.go:304-317 and returns early. An hour later it repeats. Desired capacity stays pinned at 5 indefinitely while we pay for 3 instances we don't want, logging a scaling error every cycle.

On master this converged: `setDesiredCapacity(ctx, desired)` ran unconditionally after the SIGTERM loop (scaler.go:494 on master), regardless of how many SIGTERMs actually landed. That call was the thing that made scale-in independent of SSM reachability, and dropping it is correct for the double-decrement fix but leaves nothing behind for the case where the handoff never starts.

This is the half of my earlier comment on :540 that e53674b didn't cover. That commit fixed the reporting (we now return an error instead of claiming success) and the repeated history walk, but not the underlying problem that no capacity change happens. 9fbdf9b added a fallback for exactly this shape, then scoped it to `TotalCount == 1`.

Suggestion: when `errNoReachableScaleInCandidates` comes back and the ASG has more than one instance, fall back to `setDesiredCapacity(ctx, desired)`. We know nothing is draining in that case (nothing accepted a command), so there's no decrement to double up on, and the ASG's own termination policy picks the victims. If you'd rather not hand selection back to the ASG, the alternative is recording the dispatch time and falling back only after N cooldowns with no movement in `DesiredCount`, but that needs state that survives a cold start.

scaler/asg.go
```diff
@@ -485,9 +472,17 @@ func (a *ASGDriver) GetAutoscalingActivities(ctx context.Context, nextToken *str
 	return svc.DescribeScalingActivities(ctx, input)
 }
 
+// isUserRequestedScaleIn reports whether an activity shrank the group because
+// someone asked for it: a desired-capacity change, or an instance terminated
+// with --should-decrement-desired-capacity, which is how Elastic CI Stack
+// agents leave the group after draining.
+func isUserRequestedScaleIn(cause string) bool {
```
petetomasik, line 505: Broadening the matcher is right for the in-run walk, but this function has a second consumer that wasn't revisited, and there the widening makes idle churn suppress our own scale-ins.

`GetLastScalingInAndOutActivity` is also called from the cold-start seeding in lambda/main.go:131, which isn't gated on Elastic CI mode. Its result lands in `lastScaleIn` (lambda/main.go:141-144), then in `ScaleInParams.LastEvent` (lambda/main.go:184), then in the unconditional cooldown gate at scaler.go:304-317. `findScaleIn` is `!disableScaleIn`, so with `DISABLE_SCALE_IN` unset this runs on every cold start.

The cause we now accept is the same one an Elastic CI Stack agent produces when it idles out and self-terminates. That's the exact string pinned as `want: true` at asg_test.go:128-131. Idle timeout is the stack's primary scale-down mechanism, so on any active queue this activity is always minutes old.

Trigger: an agent idles out at 10:00 and self-terminates with decrement. At 10:05 the Lambda execution environment recycles. The seed picks up the 10:00 timestamp, `LastEvent` is non-zero and recent, and the cooldown check at scaler.go:304 blocks scale-in until 11:00. We did nothing, we're being held off by someone else's idle churn, and because the fleet keeps idling out, the next cold start inherits a fresh hour again. Before this change the matcher required `userRequestForChangingDesiredCapacity`, so the seed only ever reflected scale-ins the scaler itself performed, which is the correct signal for a scaler-owned cooldown.

There's an inversion worth naming: the comment you added at scaler.go:321-326 argues the history can't tell our scale-ins apart from idle churn, which is why the walk is now gated on `LastEvent.IsZero()`. But because seeding almost always finds a self-termination, `LastEvent` is almost never zero, so the widened matcher is rarely exercised where it was intended (the walk) and constantly exercised where it hurts (the seed).

Suggestion: keep two matchers and pick per caller, e.g. an extra bool on `GetLastScalingInAndOutActivity` so lambda/main.go asks for desired-capacity causes only and the elastic walk asks for both. To head off the obvious objection: yes, the narrow matcher means seeding finds nothing on a Linux Elastic CI stack once the pre-upgrade `SetDesiredCapacity` activities age out. That's fine, and it's the flow your comment describes. `LastEvent` stays zero, the walk at scaler.go:327 fires once, and it seeds `LastEvent` at :350 so later polls use the in-process cooldown.

scaler/scaler.go
```diff
@@ -556,6 +479,49 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
...
+	}
+	if len(instanceIDs) == 0 {
+		return nil
+	}
+
+	// Every attempt starts the cooldown. Instances that accepted the command
+	// stay in the ASG until they've drained, so retrying sooner would just
+	// pick the same ones again, and a failed attempt should back off rather
+	// than retry on every poll.
+	s.scaleInParams.LastEvent = time.Now()
```
petetomasik, line 500: "Every attempt starts the cooldown" isn't quite true, and the two paths that skip it are the ones that reopen the unbounded history walk from my earlier comment on :540.

fb3c6b21 moved `LastEvent` to fire on any attempt with candidates, which fixed the main path. But `scaleIn` has two reachable exits that never get here:

1. `oldestInstances` returns an error and we bail at scaler.go:410-413, before `gracefullyScaleIn` is called at all. Trigger: `ec2:DescribeInstances` throttling. `RequestLimitExceeded` is plausible for a Lambda polling every `LAMBDA_INTERVAL` (10s default) against a large ASG, and 29d9618 deliberately made this fail rather than fall back to ASG order, so the throttle now propagates.

2. `oldestInstances` succeeds but returns an empty slice, and the guard at :506-508 returns nil before line 514. Trigger: every candidate ID stale in EC2, which `describeInstancesTolerant` drops by design and returns an empty output for (asg.go:735-738), or every instance missing a `LaunchTime` (skipped at scaler.go:640-642).

Both leave `LastEvent` zero, so the next poll re-enters the elastic block at :327, and since `MaxDescribeScalingActivitiesPages` is force-set to -1 at :329-334, we walk the entire `DescribeScalingActivities` history under the 30s budget again. Every 10 seconds, until something succeeds. That's the same cost I flagged originally, and `DescribeScalingActivities` is itself rate limited, so a throttle on one API turns into pressure on another.

Path 2 is also a silent no-op: nothing is stopped, capacity is untouched, and `scaleIn` returns nil, so `Run` reports success. Same shape as the `dispatched == 0` case, on a path e53674b didn't reach.

Suggestion: don't infer "we already walked the history" from `LastEvent`. A dedicated `historyWalked bool` on the Scaler (or a separate `lastScaleInAttempt` timestamp) decouples the two things this field is now doing, and both of these exits can set it without pretending a scale-in happened. While you're here, the guard at :506-508 reads as a precondition but runs after the call it appears to guard, and `SendSIGTERMToAgentsBatch` already short-circuits on empty input at asg.go:540-542. Moving it above the dispatch at :494 would make path 2 obvious and let it return something distinguishable from success.

scaler/asg.go
```diff
@@ -536,52 +532,98 @@ func (a *ASGDriver) GetLastScalingInAndOutActivity(ctx context.Context, findScal
...
+	// written before the stop so the drain can see it, and removed again if
+	// both stop commands fail; a stale marker would block every retry.
 	command := `#!/bin/bash
 if [ -f ` + terminationMarkerPath + ` ]; then
   echo "Already marked for termination, skipping"
   exit 0
 fi
 echo "Termination requested at $(date)" > ` + terminationMarkerPath + `
-sudo systemctl stop buildkite-agent.service || sudo /opt/buildkite-agent/bin/buildkite-agent stop --signal SIGTERM
+if ! sudo systemctl stop buildkite-agent.service && ! sudo /opt/buildkite-agent/bin/buildkite-agent stop --signal SIGTERM; then
```
petetomasik: On test coverage: the marker-clearing fix here has no test, and neither does the direct-termination gate from 9fbdf9b. Both were added in response to review findings on this PR, and both are currently free to regress.

Revert :599-602 back to the old form and every test still passes. The stop script is only ever built as a Go string and handed to a stub `SendCommand`; nothing executes it. That's a gap rather than an inherent limitation, because the sibling script already has a real harness: `TestLinuxCheckCommand` (asg_test.go:436-503) stubs `systemctl` on `PATH`, templates `terminationMarkerPath` to a temp file, and covers seven states including the `deactivating` + `ExecStopPost` window. The same pattern plus a `sudo` stub that just runs `exec "$@"` covers this script.

The regression it leaves open is permanent, not transient. If both stop commands fail and the marker survives, every later graceful-stop command hits :594-596, prints "Already marked for termination, skipping", exits 0, and is still counted at :622, so we report a successful dispatch forever. Meanwhile the check script sees marker plus live unit and reports `DRAINING` (:116-121), so `checkAndMarkUnhealthy` never marks it either. The instance is wedged with no agent stop and no cleanup, which is the class of stuck instance #350 is about.

Four cases worth pinning: marker already present (exit 0, `systemctl` never invoked); `systemctl stop` succeeds (marker retained, the `buildkite-agent stop` fallback not invoked); `systemctl stop` fails but the fallback succeeds (marker retained, exit 0); both fail (marker removed, exit 1).

Separately, and the reason the 9fbdf9b gate is also uncovered: `scaleIn` asserts on the concrete `*ASGDriver` at scaler.go:327 and :406, and no test constructs a `Scaler` with a real one, so `asgTestDriver` can never enter the graceful branch. `oldestInstances`, `gracefullyScaleIn` and `directlyTerminateInstance` are each tested in isolation, but the wiring between them isn't, including both conditions you added to the `ec2:TerminateInstances` gate. That's the highest-consequence branch in the file, since scale-in protection explicitly doesn't cover direct EC2 terminations. Worth covering: one InService plus one Terminating with `DesiredCount == 1` (rescue must be suppressed); the sentinel on a genuine single-instance ASG (rescue fires, and desired capacity drops before the terminate); an SSM API error rather than the sentinel (rescue must not fire).

Both assertions only need `Name` and `GetLastScalingInAndOutActivity`, so narrowing them to a small interface makes the branch reachable from tests and matches the interfaces-for-dependencies convention the repo already documents. One thing to fix first if you go that way: `asgTestDriver.Describe` sets `TotalCount: d.desiredCapacity` (scaler_test.go:492), but production derives it from `len(asg.Instances)` (asg.go:401). As written the mock can't represent total > InService count, which is exactly the state :423 discriminates on.

petetomasik (changes requested):

scaler/scaler.go
scadu: Fixed in aa08a82

scaler/scaler.go
scadu: Continuation in #352

scaler/asg.go
scadu: fb3c6b2 (MaxErrors 100%, counter is `accepted` with a comment), becf6af (marker cleared when both stops fail), 32c32bd (script extracted, runs under bash, TestLinuxStopCommand for the four paths).

scaler/scaler.go
scadu: Went with first option, aa08a82: nothing drainable -> setDesiredCapacity, any size. Windows takes the same path.

scaler/scaler.go
scadu: Empty-candidates exit fixed in aa08a82 (goes through the fallback). Describe-error exit disappears with the walk in #352, so no historyWalked flag.

scaler/asg.go
scadu: 32c32bd, 7ff49bc (EC2/SSM clients behind interfaces, `*ASGDriver` assertion at the graceful branch gone), aa08a82 (converge table test). The last `*ASGDriver` assertion is the walk, deleted in #352.

scaler/asg.go
scadu: Went narrow, 617fe3e in #352 (seed is its only consumer now): desired-capacity change + shrinking, terminate-with-decrement is out. One trade-off worth saying out loud: in Elastic mode on Linux our stops never carry that cause, so a cold start seeds nothing and the first poll can scale in right away. One poll, then LastEvent is in-process, and the graceful path only removes one instance's worth per pass, so I'm fine with it. Let me know if you'd like to change it.

petetomasik (approved):

PR #352, opened 2026-09-01
Persist the scale-in cooldown in SSM Parameter Store

scadu:
Fixes SUP-7940.

Since #351 the agent owns the decrement on graceful scale-in, so the ASG activity history can't tell our scale-in from an idle-timeout self-termination. A fresh Lambda container starts with no cooldown and can fire another batch straight after a recycle. Overlapping containers had the same problem.

The scaler now keeps the last scale-in time in an SSM parameter, `/buildkite-agent-scaler/<nested scaler stack name>/last-scale-in`. It reads it before choosing a direction and writes it before any graceful stop goes out. `PutParameter` returns the new version, so when two containers race only the first writer dispatches; the other backs off. A failed read or write scales nothing that poll. Candidate selection and the SSM online check happen before the write, so a throttled lookup retries next poll instead of burning the cooldown.

The template grants the permissions on the role it creates and sets the env var only then. Deployments that bring their own `AutoscalingLambdaExecutionRole` behave as before. A missing parameter counts as "never", so the first scale-in after upgrade may skip an in-flight cooldown, once per stack.

Tests cover the store, the race and the Run paths. On a sandbox stack a scale-in wrote the parameter and a forced cold start read it back instead of dispatching again.


scaler/scaler.go
```diff
@@ -317,35 +326,24 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
...
-				log.Printf("ℹ️ [Elastic CI Mode] Setting MAX_DESCRIBE_SCALING_ACTIVITIES_PAGES from %d to -1 (unlimited) for better safety checks",
-					driver.MaxDescribeScalingActivitiesPages)
-				driver.MaxDescribeScalingActivitiesPages = -1
-			}
-
-			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	if s.elasticCIMode && s.scaleInParams.CooldownPeriod > 0 {
+		// Check the ASG activity history even when LastEvent is set. Other
+		// Lambda containers scale in too, and only the history sees them.
+		if driver, ok := s.autoscaling.(*ASGDriver); ok {
```
petetomasik: fe69840, the previous commit in this stack, removed the every-poll walk deliberately and recorded why in its message:

    Idle self-terminations use the same cause text as the stops we ask for,
    so walking history every poll would extend the cooldown on unrelated
    churn. The in-run walk is now a fallback for when LastEvent is zero, and
    it seeds LastEvent so later polls use the in-process cooldown.

This commit reverses that and deletes the comment that explained it. The cross-container visibility problem it solves is real, but the mechanism over-corrects.

`isUserRequestedScaleIn` matches `"shrinking the capacity"` + `"was taken out of service in response to a user request"`, which is exactly what an Elastic CI Stack agent produces when `--disconnect-after-idle-timeout` fires and the PostStop script terminates the instance with decrement (README.md:62-68; the verbatim cause strings are pinned in asg_test.go:55).

On Elastic CI defaults (`SCALE_IN_COOLDOWN_PERIOD=1h`):

1. An agent idles out, say 5 minutes ago.
2. The next poll wanting to shrink finds that activity, sets `LastEvent` (line 348), returns nil.
3. For the next hour the early return at 313-326 short-circuits.
4. At T+1h the walk runs again. The window has slid forward, and on any queue with at least one idle-out per hour it finds another self-termination and blocks for another hour.

The cooldown renews rather than counting once, so scaler-driven scale-in never fires again in steady state. Capacity still falls via self-termination, so the fleet is not stuck, but the scaler loses the ability to make a large correction, and `SCALE_IN_COOLDOWN_PERIOD` stops meaning what the docs say it means.

Suggestion: for cooldown purposes, match only `userRequestForChangingDesiredCapacity` + `shrinkingKey`, which is what the scaler itself emits. That keeps the every-poll walk and the cross-container visibility without letting idle churn renew the window.

Separately, this behavior change is untested: re-adding `&& s.scaleInParams.LastEvent.IsZero()` to this line reverts it entirely and `go test ./scaler/` still passes.

scaler/scaler.go
```diff
@@ -317,35 +326,24 @@ func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGr
...
 			defer cancel()
 
-			// Get the last scale-in activity from ASG history
-			_, lastScaleInActivity, err := driver.GetLastScalingInAndOutActivity(ctx, false, true)
+			// Get the last scale-in activity within the configured cooldown period.
+			activitySince := time.Now().Add(-s.scaleInParams.CooldownPeriod)
+			_, lastScaleInActivity, err := driver.GetLastScalingInAndOutActivity(ctx, false, true, activitySince)
 			if err != nil {
-				log.Printf("⚠️ [Elastic CI Mode] Could not check last ASG scale-in activity: %v", err)
+				return fmt.Errorf("check last ASG scale-in activity: %w", err)
```
petetomasik: Failing closed is right, but returning here without recording anything means failure is the one outcome that cannot self-limit.

The old code failed open: on a lookup error it scaled in anyway, which set `LastEvent` and short-circuited the next hour of polls at 313-326. The loop terminated itself. Now nothing is written, so the next poll re-enters this block and calls the same API again.

Trigger: `ThrottlingException` on `DescribeScalingActivities`, Elastic CI mode, cooldown expired (or `LastEvent` still zero), ASG wants to shrink. Poll, throttled, error, no state change, poll again in `MinPollInterval`.

Volume on the shipped defaults (`template.yaml`: `MinPollInterval` 10s, `LAMBDA_TIMEOUT` 50s, 1-minute schedule) is roughly one poll every 10s continuously, and `scaler/aws_config.go:16-26` sets `RetryModeAdaptive` with `WithRetryMaxAttempts(8)`, so each poll is up to 8 signed attempts. That is up to ~48 `DescribeScalingActivities` attempts per minute per stack, against the API that is already throttling, in the exact scenario SUP-7940 describes.

Two things make it worse:

- Auto Scaling throttles on a shared per-account/region bucket across operations, so this also starves `DescribeAutoScalingGroups` and `SetDesiredCapacity` for every other queue's scaler in the account. A customer with 20 queues has 20 stacks entering the loop at once.
- `scaler/asg.go:473-479` builds a fresh `autoscaling.NewFromConfig(a.Cfg)` per call, so the adaptive rate limiter learns the throttle rate across one call's 8 attempts and then forgets it. The only backoff surviving between polls is the 10s interval.

scaler/asg.go
```diff
@@ -482,51 +488,69 @@ func isUserRequestedScaleIn(cause string) bool {
...
+		})
+	}
 
-	for i := 0; !hasFoundScalingActivities; {
-		i++
-		if a.MaxDescribeScalingActivitiesPages >= 0 && i >= a.MaxDescribeScalingActivitiesPages {
-			return lastScalingOutActivity, lastScalingInActivity, fmt.Errorf("%d exceeds allowed pages for autoscaling:DescribeScalingActivities, %d", i, a.MaxDescribeScalingActivitiesPages)
+	for page := 0; ; page++ {
+		if a.MaxDescribeScalingActivitiesPages > 0 && page >= a.MaxDescribeScalingActivitiesPages {
+			return lastScalingOutActivity, lastScalingInActivity, fmt.Errorf("autoscaling:DescribeScalingActivities exceeded maximum of %d pages", a.MaxDescribeScalingActivitiesPages)
```
petetomasik: Two problems meet here, and together they can silently stop scale-in.

First, the semantics: exhausting the page budget means "no scale-in found in the window yet", not "the lookup failed". Those are different facts and they want different handling. Returning a plain error conflates them, and `scaleIn` now treats any error as fail-closed.

Second, this PR removes the guard that used to prevent the collision. The old Elastic CI block forced `driver.MaxDescribeScalingActivitiesPages = -1` before walking the history, specifically so a user-set cap could not truncate the safety check. Nothing replaces it for the Lambda path, which passes the env value straight through at `lambda/main.go:202`.

Trigger: `ELASTIC_CI_MODE=true`, `SCALE_IN_COOLDOWN_PERIOD=1h`, `MAX_DESCRIBE_SCALING_ACTIVITIES_PAGES=2`, and more than 200 successful activities in the window with none matching `isUserRequestedScaleIn`. A ramping queue does this on its own, since every instance launch is a successful activity. The cap fires, `scaleIn` returns the error, and scale-in is blocked on every poll for as long as the churn lasts. There is no CloudWatch signal: `lambda/main.go:216-218` logs and continues, so no `Errors` metric and no DLQ message.

Note the shape of it. The budget can only be exhausted on the branch where no recent scale-in was found, which is exactly the case where scaling in is safe. So the cap inverts from "bound the scan" into "refuse to scale in".

The default is `-1` and elastic-ci-stack-for-aws does not set the parameter, so this only reaches hand-configured stacks, and that population overlaps heavily with operators who were already tuning around throttling. The parameter is an unconstrained `Number` in `template.yaml:134-137` with no `MinValue`.

scaler/scaler_test.go
```diff
@@ -830,6 +852,89 @@ func TestScalerRunWaitsToScaleInWhileInstancesArePending(t *testing.T) {
...
+		t.Errorf("scaleIn() error = %v, want %v", err, context.Canceled)
+	}
+}
+
+func TestScaleInHonorsRecoveredActivity(t *testing.T) {
+	client := &stubScalingActivitiesClient{
+		responses: []stubScalingActivitiesResponse{{
+			output: &autoscaling.DescribeScalingActivitiesOutput{
+				Activities: []types.Activity{{
+					Cause:      aws.String("At 2026-08-26T16:14:36Z instance i-017bacd33066f0408 was taken out of service in response to a user request, shrinking the capacity from 2 to 1."),
```
petetomasik: The behavior change this PR's second commit is built around has no test.

All three new `scaleIn` tests construct `Scaler{scaleInParams: ScaleParams{CooldownPeriod: time.Hour}}` and leave `LastEvent` at its zero value, so every one of them takes the old cold-start path. The warm-container case the commit message describes ("other containers scale in too, and the ASG activity history is the only place that shows them") is never exercised.

Verified by mutation: re-adding `&& s.scaleInParams.LastEvent.IsZero()` to scaler.go:332 reverts the entire change and `go test ./scaler/` stays green.

The missing case: set `LastEvent = time.Now().Add(-2 * time.Hour)` against `CooldownPeriod: time.Hour` so the local cooldown check at scaler.go:313-326 falls through, have the stub return one Successful activity whose cause contains `"was taken out of service in response to a user request"` and `"shrinking the capacity"` with `StartTime = time.Now().Add(-time.Minute)`, then assert both that `scaleIn` returns nil and that `s.scaleInParams.LastEvent` advanced to that `StartTime`.

Worth writing that one first, because it is also where the cooldown-renewal question I raised on scaler.go:332 becomes concrete. Whatever the test asserts is the answer to what `SCALE_IN_COOLDOWN_PERIOD` should mean when an instance self-terminates.

While here: this test asserts only `err == nil`, which is also what a successful scale-in returns. It distinguishes the two cases only by accident, since falling through would reach `SetDesiredCapacity` on an `ASGDriver` holding a zero `aws.Config` and the SDK's missing-region error would trip the `t.Fatalf`. That is an environment-dependent oracle and it stops working the moment the test grows a stubbed ASG client. Asserting no `SetDesiredCapacity` call, plus `LastEvent` equal to the activity's `StartTime`, would make the name and the assertion agree.

petetomasik (changes requested):

scadu:
@petetomasik, I rebuilt this on #351's current head, so most of what you reviewed is gone. fe649c3 deletes the per-poll walk (plus `scaleInHistoryChecked` and the two matchers). Cooldown is in-process `LastEvent`, seeded once per Lambda lifetime from ASG history bounded by `StartTimeLowerBound = now - max(cooldowns)` and `Status=Successful` (5d3a0d3). 
Lookup failure seeds `lastScaleIn = now`, one cooldown instead of a retry loop (98cdeb4). Seed matcher narrowed to desired-capacity changes (617fe3e). fdd15c6 makes Elastic mode honour `DISABLE_SCALE_IN`.

scaler/scaler_test.go
scadu: Gone with the walk. Seed is covered by `TestSeedLastScaleTimes` plus the `DescribeScalingActivities` stub tests in asg_test.go (filters, since bound, pagination, no-match, page limit, API error, skip). No zero-config ASGDriver in scaleIn tests anymore.

scaler/asg.go
scadu: Partially addressed in 5d3a0d3 bounds the walk by time, so it's a page or two even on a busy ASG, and it runs once per cold start, so a low `MAX_DESCRIBE_SCALING_ACTIVITIES_PAGES` can't block every poll anymore.

scaler/scaler.go
scadu: Sidestepped: lookup runs once per container, and on failure we seed `lastScaleIn = now` (98cdeb4). One throttle = one cooldown, then zero `DescribeScalingActivities` calls from that container. Fresh client per call is still there but only for that one call; can hoist it if you care.

scaler/scaler.go
```diff
@@ -296,16 +296,10 @@ func (s *Scaler) Run(ctx context.Context) (time.Duration, error) {
 }
 
 func (s *Scaler) scaleIn(ctx context.Context, desired int64, current AutoscaleGroupDetails) error {
-	// In ElasticCIMode, DISABLE_SCALE_IN is ignored (handled by s.elasticCIMode check below)
-	if s.scaleInParams.Disable && !s.elasticCIMode {
+	if s.scaleInParams.Disable {
```
petetomasik, line 299: This drops the `&& !s.elasticCIMode`, which makes `DISABLE_SCALE_IN` authoritative in Elastic CI Mode. Combined with the template defaults, that turns off graceful scale-in for the default deployment.

`template.yaml:117` defaults `DisableScaleIn` to `"true"` and `template.yaml:346` wires it straight to `DISABLE_SCALE_IN`. I checked the consumer rather than assuming: `elastic-ci-stack-for-aws` `templates/aws-stack.yml` also declares `DisableScaleIn` with `Default: "true"`, and passes `DisableScaleIn: !Ref DisableScaleIn` alongside `EnableElasticCIMode: !Ref ScalerEnableExperimentalElasticCIMode` to both scaler nested stacks (the x86-64 and arm64 `AWS::Serverless::Application` resources).

Trigger: flip `ScalerEnableExperimentalElasticCIMode` to `true` and change nothing else. Every poll now returns nil here, so there is no graceful stop, no desired-capacity change, and no `terminateIfAgentUnreachable`. The `SendSIGTERMToAgentsBatch` / `oldestInstances` machinery from #345, #346 and #351 is unreachable, and instances only ever leave via `--disconnect-after-idle-timeout`.

It also makes this PR's own subject inert at defaults. `disableScaleIn == true` means `findScaleIn = false` at `lambda/main.go:170`, so `lastScalingActivities` never looks for a scale-in, and `seedLastScaleTimes` skips the fail-closed seed entirely. The cooldown recovery only does anything for operators who have explicitly set `DisableScaleIn=false`.

And `docs/elastic_ci_mode.md:12`, in a file this PR edits, still says `ELASTIC_CI_MODE=true` "enables graceful scale-in for Elastic CI Stack". At template defaults that is now false.

Reversing a decision that shipped with Elastic CI Mode itself is a big behaviour change to carry in a PR titled "Recover Elastic CI scale-in cooldown from ASG activity". Since `RELEASE.md` generates notes with `ghch` from merged PR titles, it would land in v1.14.0 under an unrelated heading. Suggest splitting fdd15c6 into its own PR, and pairing it with either a flipped default here, a `Condition` that forces `DISABLE_SCALE_IN=false` when `EnableElasticCIMode` is true, or a coordinated change on the stack side. Whichever way, `template.yaml:112` and `docs/elastic_ci_mode.md:12` need to be reconciled with it.

petetomasik, line 299: Second consequence of making `DISABLE_SCALE_IN` authoritative here: it also switches off the dangling-instance rescue for a lone instance whose agent has died, and pushes dangling cleanup out to the 1h uptime gate.

`Run` returns via this function at `scaler/scaler.go:272` whenever `asg.DesiredCount > desired`, so the 10-minute zero-agents branch at `scaler/scaler.go:284` is only reached when `desired == asg.DesiredCount`. That call is now a no-op when `Disable` is set, so nothing runs on either path.

Trigger: Elastic CI Mode, `DISABLE_SCALE_IN=true` (the template default, see my other comment), `MinSize=0`, ASG at `DesiredCount=1, TotalCount=1`, agent process dead so `metrics.TotalAgents=0` and no jobs so `desired=0`. Every poll takes the `0 < 1` branch at line 263, returns here immediately, and never evaluates line 284. Cleanup falls back to the top-of-`Run` call at `scaler/scaler.go:161`, which is gated on `DANGLING_CHECK_MINIMUM_INSTANCE_UPTIME`, default 1h. Before fdd15c6 this instance was rescued within one poll: `scaleIn` ran the elastic branch, `gracefullyScaleIn` failed with `errNoReachableScaleInCandidates`, and `terminateIfAgentUnreachable` at `scaler/scaler.go:402` terminated it directly.

Worth noting the comment at `scaler/scaler.go:275-283` states the assumption this breaks in as many words: "neither scaleIn() nor scaleOut() is called because desired == instanceCount after MaxSize capping". That was true when `scaleIn` always did something in Elastic CI Mode. It no longer is, so the branch that exists specifically to catch dead agents is now unreachable in the case where capacity is above what the metrics want, which is the common shape for a dead agent.

Two options: evaluate the zero-agents dangling branch before the scale-in early return, or gate the return at line 272 so line 284 is still reached when scale-in is disabled. Either way this wants a test, since `TestScaleInDisabledInElasticCIMode` pins the new `Disable` behaviour but nothing covers `Run` reaching the dangling branch with `Disable=true`.

scaler/asg.go
```diff
@@ -489,70 +488,92 @@ func (a *ASGDriver) SetDesiredCapacity(ctx context.Context, count int64) error {
...
+// a desired-capacity change, which is how the scaler scales in outside
+// Elastic CI Mode. Instances that terminate with
+// --should-decrement-desired-capacity are deliberately not matched: an agent
+// idling out writes the same cause as one we asked to stop, so treating them
+// as scale-ins would make an idle-out restart the cooldown.
 func isUserRequestedScaleIn(cause string) bool {
 	return strings.Contains(cause, shrinkingKey) &&
-		(strings.Contains(cause, userRequestForChangingDesiredCapacity) ||
-			strings.Contains(cause, userRequestForTerminatingInstance))
+		strings.Contains(cause, userRequestForChangingDesiredCapacity)
```
petetomasik, line 505: This narrowing means the Elastic CI graceful path leaves no activity the cold-start seed can ever match, so the cooldown still does not survive a container recycle. buildsworth's blocking comment is correct on 617fe3e.

The trace:

1. On the graceful happy path `scaler/scaler.go:369` returns without calling `setDesiredCapacity`, deliberately, because the agent owns the decrement.
2. The drained agent decrements itself, and AWS writes the cause `"... was taken out of service in response to a user request, shrinking the capacity from N to N-1."`
3. That cause does not contain `userRequestForChangingDesiredCapacity`, so this predicate no longer matches it. Pinned by the test case you changed at `scaler/asg_test.go:329-334`.
4. Dangling cleanup uses `SetInstanceHealth` (`scaler/asg.go:394`), which replaces rather than shrinks, so it produces nothing either.

Net effect: on a healthy Linux Elastic CI stack, a successful graceful scale-in leaves zero matchable activity. A fresh container seeds `lastScaleIn` as zero, and because the cooldown check at `scaler/scaler.go:304` is gated on `!s.scaleInParams.LastEvent.IsZero()`, it is skipped outright rather than treated as "no scale-in yet". Container A dispatches a batch at T, gets recycled at T+10m, container B dispatches another full batch with 50 minutes of the documented `SCALE_IN_COOLDOWN_PERIOD` outstanding.

Correcting one thing in my earlier comment: this is a cross-container gap only, not a per-poll one. `gracefullyScaleIn` does set `LastEvent` at `scaler/scaler.go:447`, and `lambda/main.go:250-251` carries it across warm invocations, so within a container the cooldown holds. Blast radius is also partly self-limiting, since the stop script's marker file makes a repeat SIGTERM to an already-draining instance a no-op.

On the substance, your reasoning for 617fe3e is sound and so was my objection to fe69840. Both are correct, which is the actual finding: the two shrink causes are byte-identical, so matching gives a cooldown that idle churn renews forever, and not matching gives no cooldown at all across containers. There is no version of this predicate that answers both. ASG activity history is the wrong signal for this mode.

The scaler already knows the answer at `scaler/scaler.go:447`. It just discards it when the container dies. Persisting it is cheap here given `scaler/ssm.go` and the SSM grant at `template.yaml:264-270` already exist: one parameter per ASG, written on scale-in and read in place of the seed block at `lambda/main.go:161-185`. That also lets `findScaleIn` go to `false`, which fixes the separate problem that this predicate never matching means the "bounded" walk never hits its early return at `scaler/asg.go:567` and pages the entire cooldown window on every cold start.

scaler/scaler.go
scadu: Yeah, you're right. Both shrink causes look the same, so there's no predicate that works. I dropped the walk from scaleIn in 8339c59. Every scale-in now goes through recordScaleIn which writes the timestamp to SSM, and a fresh container reads it once at cold start.

scaler/scaler.go
scadu: Ignore 98cdeb4 from my earlier reply, that commit's gone, but the idea survived. The SSM read happens once per container and if it fails we set `lastScaleIn = now`. So a bad read costs one cooldown and that's it, no more SSM calls from that container. There's no `DescribeScalingActivities` on the scale-in path at all now. Scale-out still walks once per cold start and fails open, same as master.

scaler/asg.go
scadu: Same story here, 5d3a0d3 is gone. I removed the scale-in walk entirely instead, so `MAX_DESCRIBE_SCALING_ACTIVITIES_PAGES` can't block scale-in anymore. The scale-out walker is untouched master code, still page-capped, still fails open. Telling "ran out of pages" apart from "lookup failed" is a fair point but I'd rather not grow this PR again, so I'll pick it up in a follow-up.

scaler/asg.go
scadu: Went with this. `recordScaleIn` sits on the three original `LastEvent` writes `/buildkite-agent-scaler/${AgentAutoScaleGroup}/last-scale-in`; the Lambda reads it once at cold start instead of the old seed block, and `template.yaml` grants `Get/PutParameter` on just that ARN. `isUserRequestedScaleIn` and the walk are gone with it.

scaler/asg.go
scadu: The in-scaler walk is gone, so the page cap and the -1 override no longer collide. The handler seed honors the cap same as master, and with the parameter set the scale-in half of that seed is skipped entirely.

scaler/asg.go
scadu: Went with the SSM parameter. The scaler records the time it dispatched a graceful stop and reads it before choosing a direction.

scaler/scaler.go
scadu: Not changing `DISABLE_SCALE_IN` behavior in this PR, so closing this thread.

PR #359, opened 2026-09-23
Enable cross region SSM access

jodelasur:
Adds an optional `BUILDKITE_AGENT_TOKEN_SSM_REGION` (flag `-agent-token-ssm-region`, CFN param `BuildkiteAgentTokenParameterRegion`) to read the agent token from a different region than the Lambda runs in.

We plan to share one SSM parameter across regions so it can be rotated in one place. Since `GetParameter` routes by the client's region — a full ARN doesn't re-route it — scalers currently can't read a parameter in another region without duplicating it per region.

When unset, behaviour is unchanged. Includes tests and README updates.


jodelasur:
Sorry, prematurely created this PR. Will open a new one once ready.

